PR #2, opened 2016-03-08
[IMP] Added validator to 'country' field

LeGaumaisMasque:
Implemented S0016: country is mandatory

### DIFF
--- a/beesdoo_base/models/partner.py
+++ b/beesdoo_base/models/partner.py
@@ -49,19 +49,3 @@ class Partner(models.Model):
                 if command[0] == 2:
                     command[0] = 3
         return super(Partner, self).write(values)
-
-class MemberCard(models.Model):
-
-    def _get_current_user(self):
-        return self.env.user
-
-    _name = 'member.card'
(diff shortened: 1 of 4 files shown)


beesdoo_base/models/partner.py
```diff
@@ -26,12 +26,20 @@ def _on_change_name(self):
         self.name = concat_names(self.first_name, self.last_name)
 
     @api.one
+    @api.constrains('country_id')
+    def _check_country(self):
+        if len(self.country_id) == 0:
+            raise ValidationError(_('Country is mandatory'))
+
```
tfrancoi: Just add required on the field country_id in the field description

LeGaumaisMasque: Je peux redéfinir country_id, genre:
`country_id = fields.Many2one('res.country', string='Country', required=True)`
?

LeGaumaisMasque: I've update the implementation (as described in the previous comment).

PR #3, opened 2016-03-21
S001

tfrancoi:


beesdoo_base/models/partner.py
```diff
@@ -49,19 +51,15 @@ def write(self, values):
                 if command[0] == 2:
                     command[0] = 3
         return super(Partner, self).write(values)
+    
+    @api.multi
+    def _deactivate_active_cards(self):
+        if len(self.member_card_ids) > 0:
```
tfrancoi, line 57: if is useless since the for will not iterate if there is no element but will not crash

beesdoo_base/models/partner.py
```diff
@@ -49,19 +51,15 @@ def write(self, values):
...
+            for c in self.member_card_ids:
+                if c.valid:
+                    c.valid = False
+                    c.end_date = fields.Date.today()      
+    @api.multi            
+    def _new_card(self, txt):
+        self.env['member.card'].create({'partner_id' : self.env.context['active_id'],'comment' : txt})
```
tfrancoi, line 64: can use self.env['member.card'].create({'partner_id' : self.env.context['active_id'],'comment' : txt})
self.id is already the record of the partner don't need to use self.env.context['active_id']

if it's @api.multi you need to iterate on self can use @api.one instead

beesdoo_base/models/partner.py
```diff
@@ -49,19 +51,15 @@ def write(self, values):
                 if command[0] == 2:
                     command[0] = 3
         return super(Partner, self).write(values)
+    
+    @api.multi
+    def _deactivate_active_cards(self):
+        if len(self.member_card_ids) > 0:
+            for c in self.member_card_ids:
```
tfrancoi, line 58: Use a better name then c, card for example

beesdoo_base/models/membercard.py
```diff
@@ -0,0 +1,47 @@
...
+    end_date = fields.Date(readonly=True, string="Date d'expiration")
+    comment = fields.Char("Raison", required=True)
+    
+# A transient model for the creation of a new card. The user can only define the raison why 
+# a new card is needed and the eater/worker that is concerned.
+class MemberCardWizard(models.TransientModel):
+    _name = 'membercard.wizard'
+   
+    new_comment = fields.Char('Raison', required=True)
+
```
tfrancoi, line 41: add partner_id field with default value self.env.context['active_id']

beesdoo_base/models/membercard.py
```diff
@@ -0,0 +1,47 @@
...
+# A transient model for the creation of a new card. The user can only define the raison why 
+# a new card is needed and the eater/worker that is concerned.
+class MemberCardWizard(models.TransientModel):
+    _name = 'membercard.wizard'
+   
+    new_comment = fields.Char('Raison', required=True)
+    
+    @api.multi
+    def create_new_card(self):
+        client = self.env['res.partner'].search([('id', '=',self.env.context['active_id'])])
```
tfrancoi, line 44: use the new field partner_id instead

need to use sudo in order to bypass the access right that is only readonly for everyone (which is good) 

``` python
    @api.one
    def create_new_card(self):
        client = self.partner_id.sudo()
        client._deactivate_active_cards()
        client._new_card(self.new_comment, self.env.uid)
```

beesdoo_base/models/membercard.py
```diff
@@ -0,0 +1,47 @@
...
+from random import randint
+import uuid
+
+class MemberCard(models.Model):
+    
+    def _get_current_user(self):
+        return self.env.uid
+    
+    def _get_current_client(self):
+        # TODO : this does not work
```
tfrancoi, line 12: It's useless since this value has to be set by the wizard

beesdoo_base/models/partner.py
```diff
@@ -10,15 +10,15 @@ class Partner(models.Model):
     first_name = fields.Char('First Name')
     last_name = fields.Char('Last Name', required=True)
 
-    eater = fields.Selection([('eater', 'Eater'), ('worker_eater', 'Worker and Eater')], string="Eater/Worker")
+    eater = fields.Selection([('eater', 'Mangeur'), ('worker_eater', 'Mangeur et Travailleur')], string="Mangeur/Travailleur")
```
tfrancoi, line 13: Keep english wording, need to add translation instead

beesdoo_base/models/partner.py
```diff
@@ -10,15 +10,15 @@ class Partner(models.Model):
...
 
     child_eater_ids = fields.One2many("res.partner", "parent_eater_id", domain=[('customer', '=', True),
                                                                                 ('eater', '=', 'eater')])
 
-    parent_eater_id = fields.Many2one("res.partner", string="Parent Worker", readonly=True)
+    parent_eater_id = fields.Many2one("res.partner", string="Parent Travailleur", readonly=True)
```
tfrancoi, line 18: Same remark

beesdoo_base/views/partner.xml
```diff
@@ -1,43 +1,74 @@
...
+					<button type="object" name="create_new_card" string="Créer"
+						class="oe_highlight" />
+					<button special="cancel" string="Annuler" />
+				</footer>
+			</form>
+		</field>
+	</record>
+
+	<act_window id="action_membercard_wizard"
+		name="Créer une nouvelle carte de membre - Désactiver l'ancienne"
```
tfrancoi, line 22: English wording + translation

beesdoo_product/__openerp__.py
```diff
@@ -20,7 +20,7 @@
     'version': '0.1',
 
     # any module necessary for this one to work correctly
-    'depends': ['beesdoo_base', 'product'],
+    'depends': ['beesdoo_base', 'product', 'point_of_sale'],
```
tfrancoi, line 23: Why this dependency is needed in this module ?

beesdoo_base/views/partner.xml
```diff
@@ -1,43 +1,74 @@
...
+				<field name="parent_eater_id" attrs="{'invisible' : [('eater', '!=', 'eater')]}" />
+			</field>
+			<xpath expr="//notebook" position="inside">
+				<page string="Carte de Membre"
+					attrs="{'invisible': ['|', ('customer', '=', False), ('eater', '!=', 'worker_eater')]}">
+					<separator string="Mangeurs" />
+					<field name="child_eater_ids" widget="many2many_tags" />
+					<separator string="Cartes" />
+					<field string="Cartes" name="member_card_ids">
+						<tree editable="bottom">
```
tfrancoi: field should not be editable anymore

beesdoo_base/models/membercard.py
```diff
@@ -0,0 +1,47 @@
...
+        return code
+    
+    _name = 'member.card'
+    _order = 'activation_date desc'
+    
+    valid = fields.Boolean(default=True, string="Active")
+    barcode = fields.Char("Code barre", oldname='ean13', default=_compute_bar_code)
+    partner_id = fields.Many2one('res.partner') #, default=_get_current_client)
+    responsible_id = fields.Many2one('res.users', default=_get_current_user, string="Responsable")
+    activation_date = fields.Date(default=fields.Date.today, readonly=True, string="Date de création")
```
tfrancoi, line 31: Use create_date which already define on all object

beesdoo_base/models/membercard.py
```diff
@@ -0,0 +1,47 @@
...
+        # TODO : this does not work
+        return self.env['res.partner'].search([('id', '=',self.env.context['active_id'])])
+    
+    def _compute_bar_code(self):
+        rule = self.env['barcode.rule'].search([('name', '=', 'Customer Barcodes')])[0]
+        nomenclature = self.env['barcode.nomenclature']
+        size = 13-len(rule.pattern)
+        ean = rule.pattern + str(uuid.uuid4().fields[-1])[:size] #str(randint(10**(size-1), 10**size-1))
+        code = ean[0:12] + str(nomenclature.ean_checksum(ean))
+        nomenclature.check_encoding(code,'ean13')
```
tfrancoi, line 21: This line return true or false, there is no point to call it, if the result is not check. 
There is no point to check it in real life once we know our code generate valid barcode. 
If we check the result and raise an error the end user cannot do anything anyway.

beesdoo_base/models/membercard.py
```diff
@@ -0,0 +1,47 @@
...
+    def _get_current_client(self):
+        # TODO : this does not work
+        return self.env['res.partner'].search([('id', '=',self.env.context['active_id'])])
+    
+    def _compute_bar_code(self):
+        rule = self.env['barcode.rule'].search([('name', '=', 'Customer Barcodes')])[0]
+        nomenclature = self.env['barcode.nomenclature']
+        size = 13-len(rule.pattern)
+        ean = rule.pattern + str(uuid.uuid4().fields[-1])[:size] #str(randint(10**(size-1), 10**size-1))
+        code = ean[0:12] + str(nomenclature.ean_checksum(ean))
```
tfrancoi: 0:12 should be useless otherwise we have a problem since we are making a checksum on the full ean..

beesdoo_base/models/membercard.py
```diff
@@ -0,0 +1,47 @@
...
+    activation_date = fields.Date(default=fields.Date.today, readonly=True, string="Date de création")
+    end_date = fields.Date(readonly=True, string="Date d'expiration")
+    comment = fields.Char("Raison", required=True)
+    
+# A transient model for the creation of a new card. The user can only define the raison why 
+# a new card is needed and the eater/worker that is concerned.
+class MemberCardWizard(models.TransientModel):
+    _name = 'membercard.wizard'
+   
+    new_comment = fields.Char('Raison', required=True)
```
tfrancoi, line 40: Use a text field instead of a char it's more confortable for the user

beesdoo_base/models/membercard.py
```diff
@@ -0,0 +1,47 @@
...
+    barcode = fields.Char("Code barre", oldname='ean13', default=_compute_bar_code)
+    partner_id = fields.Many2one('res.partner') #, default=_get_current_client)
+    responsible_id = fields.Many2one('res.users', default=_get_current_user, string="Responsable")
+    activation_date = fields.Date(default=fields.Date.today, readonly=True, string="Date de création")
+    end_date = fields.Date(readonly=True, string="Date d'expiration")
+    comment = fields.Char("Raison", required=True)
+    
+# A transient model for the creation of a new card. The user can only define the raison why 
+# a new card is needed and the eater/worker that is concerned.
+class MemberCardWizard(models.TransientModel):
```
tfrancoi, line 37: Put the wizard in wizard folder

beesdoo_base/views/partner.xml
```diff
@@ -1,43 +1,74 @@
...
-            <field name="barcode" position="attributes">
-                <attribute name="attrs">{'invisible' : [('eater', '!=', 'worker_eater')]}</attribute>
-            </field>
-            <field name="barcode" position="after">
-                <field name="parent_barcode" attrs="{'invisible' : [('eater', '!=', 'eater')]}" />
-            </field>
-        </field>
-    </record>
+	<record id="MemberCard Wizard" model="ir.ui.view">
+		<field name="name">New MemberCard Wizard</field>
```
tfrancoi, line 5: Change the view to adapt to text fields

``` xml
<separator string="Reason" />
<field name="new_comment" string="Raison" editable="True" />
<field name="partner_id" invisible="1" />
<footer>
     <button type="object" name="create_new_card" string="Créer"
                       class="oe_highlight" />
    <button special="cancel" string="Annuler" />
</footer>
```

beesdoo_base/views/partner.xml
```diff
@@ -1,43 +1,74 @@
...
+				<footer>
+					<button type="object" name="create_new_card" string="Créer"
+						class="oe_highlight" />
+					<button special="cancel" string="Annuler" />
+				</footer>
+			</form>
+		</field>
+	</record>
+
+	<act_window id="action_membercard_wizard"
```
tfrancoi, line 21: No need of an act window of this form (this tag create also ir.values that we don't need here) 

``` xml
    <record model="ir.actions.act_window" id="action_membercard_wizard">
        <field name="name">New Member Card</field>
        <field name="res_model">membercard.new.wizard</field>
        <field name="view_mode">form</field>
        <field name="target">new</field>
    </record>
```

beesdoo_base/models/partner.py
tfrancoi, line 64: Need a parameter to get the user_id since this method is called with sudo and we loose the information of the user
    @api.multi
    def _new_card(self, reason, user_id):
        self.env['member.card'].create({'partner_id' : self.id,'responsible_id' : user_id, 'comment' : reason})

beesdoo_purchase/views/purchase_order.xml
```diff
@@ -0,0 +1,14 @@
+<openerp>
+	<data>
+		<record model="ir.ui.view" id="beesdoo_purchase_order_form_view">
+			<field name="name">beesdoo.purchase.order.form.view</field>
+			<field name="model">purchase.order</field>
+			<field name="inherit_id" ref="purchase.purchase_order_form" />
+			<field name="arch" type="xml">
+				<field name="product_id" position="attributes">
```
tfrancoi, line 8: It's not correct:
                <field name="product_id" position="attributes">
                    <attribute name="domain">[('main_seller_id','=', parent.partner_id)]</attribute>
                </field>

inside it's attribute without S then name="domain" and then the parent.partner_id should not be quoted

beesdoo_purchase/views/purchase_order.xml
```diff
@@ -0,0 +1,14 @@
...
+					<attributes domain="[('main_seller_id','=','parent.partner_id')]" />
+				</field>
+			</field>
```
tfrancoi, line 11: Correct Version 
                <field name="product_id" position="attributes">
                    <attribute name="domain">[('main_seller_id','=', parent.partner_id)]</attribute>
                </field>

PR #4, opened 2016-04-03
feat: make coin_value field of account.bnk_stmt_cashbox.form read-only

leeuwerck:


beesdoo_pos/views/beesdoo_pos.xml
```diff
@@ -16,6 +16,17 @@
...
         </record>
 
+         <record id="view_account_bnk_stmt_cashbox" model="ir.ui.view">
+            <field name="name">account.bnk_stmt_cashbox.form</field>
+            <field name="model">account.bank.statement.cashbox</field>
+             <field name="inherit_id" ref="account.view_account_bnk_stmt_cashbox" />
+            <field name="arch" type="xml">
+               <field name="coin_value" position="replace">
+                   <field name="coin_value" readonly="1"/>
+               </field>
```
tfrancoi, line 26: use attribute instead of replace 

``` xml
<field name="coin_value" position="attributes">
                <attribute name="readonly">1</attribute>
            </field>
```

PR #13, opened 2016-08-12
9.0 s010cr elise new

tfrancoi:


beesdoo_base/models/partner.py
```diff
@@ -74,4 +74,16 @@ def _auto_init(self, cr, context=None):
...
+    @api.multi
+    def _new_eater(self, surname, name, email):
+        partner_data = {
+                        'name' : surname + " "+ name,
+                        'last_name' : name,
+                        'first_name' : surname,
+                        'is_customer' : True,
+                        'eater' : 'eater',
+                        'parent_eater_id' : self.id,
+                        'parent_id' : self.id,
```
tfrancoi, line 86: Pas sur que dans le cas des part A le partner doit être parent_id (vu que c'est pour la relation compagnie <=> personne) 

Il faudra surement tenir compte de la différence part A et B

beesdoo_base/wizard/partner.py
```diff
@@ -0,0 +1,22 @@
...
+
+    first_name = fields.Char('First Name', required=True)
+    last_name = fields.Char('Last Name', required=True)
+    email = fields.Char('Email')
+    
+    partner_id = fields.Many2one('res.partner', default=_get_default_partner)
+    
+    @api.one
+    def create_new_eater(self):
+        client = self.partner_id.sudo()
```
tfrancoi, line 21: In this case the sudo is not needed. In the cas of the beescard, we don't want regular user to be able to create card without the wizard. But in this case the current user should be able to create a partner, it's just a short cut

beesdoo_base/models/partner.py
tfrancoi, line 86: I've Added country_id since it's required now

PR #26, opened 2017-10-28
[IMP] website_shift: Change rules of display and design

remytms:
The menu entry was recorded in data folder. But this folder is reserved
for buisness data. So I moved the menu entry in the views directory.

[IMP] website_shift: Responsive design

[IMP] website_shift: change rules of display

Shows only available space:
    if more than 5 max worker and at least two available spaces
    if less than 5 max worker and at least one available spaces

This take care of the case where no task_template is defined in the task

[IMP] website_shift: Make template CMS compliant


remytms:
This PR is already tested by @vdewulf.

remytms:
BEES coop need this PR before his 'Assemblée Générale' on decembre the 3th.

Can someone check my code ? :)

tfrancoi:
Merged

PR #33, opened 2018-01-10
[ADD] website_shift: personal page for working information

remytms:
This PR adds a personal page for each user that shows usefull information about the work of the cooperator in the cooperative.
This allows irregular worker to subscribe to a shift via the website.

I have also refactored all the templates and the controller, could you tell me if it's better ? :)

I still have to add some little functionnalities, but that doesn't change a lot.

Thanks for your review. :)

### DIFF
--- a/beesdoo_website_shift/controllers/main.py
+++ b/beesdoo_website_shift/controllers/main.py
@@ -16,6 +16,8 @@ class WebsiteShiftController(http.Controller):
             return self.shift_irregular_worker()
         if working_mode == 'regular':
             return self.shift_template_regular_worker()
+        if working_mode == 'exempt':
+            return self.shift_exempted_worker()
 
         return request.render(
             'beesdoo_website_shift.shift',
@@ -122,5 +124,17 @@ class WebsiteShiftController(http.Controller):
                 'task_templates': task_templates,
                 'float_to_time': float_to_time,
                 'subscribed_shifts': subscribed_shifts,
+            }
+        )
+
+    def shift_exempted_worker(self, **kwargs):
+        # Get current user
+        cur_user = request.env['res.users'].browse(request.uid)
+
+        return request.render(
+            'beesdoo_website_shift.exempted_worker',
+            {
+                'partner': cur_user.partner_id,
+                'status': cur_user.partner_id.cooperative_status_ids,
             }
         )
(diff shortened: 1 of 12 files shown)


beesdoo_website_shift/controllers/main.py
```diff
@@ -1,56 +1,437 @@
...
-from openerp import http
+
+from openerp import http, fields
 from openerp.http import request
+from openerp.tools import DEFAULT_SERVER_DATETIME_FORMAT as DATETIME_FORMAT
 
 from openerp.addons.beesdoo_shift.models.planning import float_to_time
 
-class ShiftPortalController(http.Controller):
+PERIOD = 28  # TODO: use the same constant as in 'beesdoo_shift'
```
remytms, line 18: I think PERIOD should be in the module beesdoo_shift. But we can do the change later.

beesdoo_website_shift/controllers/main.py
```diff
@@ -1,56 +1,437 @@
...
+        template_context.update(self.my_shift_worker_status())
+        template_context.update(self.my_shift_next_shifts())
+        template_context.update(self.my_shift_past_shifts())
+        template_context.update(self.available_shift_irregular_worker(
+            irregular_enable_sign_up, nexturl
+        ))
+
+        future_alert_date = False
+        if not cur_cooperative_status.alert_start_time:
+            # Compute date before which the worker is up to date
```
remytms, line 149: Maybe future_alert_date can be computed in the cooperative status directly.

beesdoo_website_shift/controllers/main.py
```diff
@@ -1,56 +1,437 @@
...
+        """
+        Return template variables for 'beesdoo_website_shift.my_shift_worker_status_*' template
+        """
+        cur_user = request.env['res.users'].browse(request.uid)
+        return {
+            'status': cur_user.partner_id.cooperative_status_ids,
+        }
+
+
+class Shift(object):
```
remytms, line 355: This class was added has I don't find a way to create records of beesdoo.shift.shift that are not saved in the database.

Maybe formatting the date in this class isn't a good idea as it can be performed in the template. But is simplify the template code.

tfrancoi: you can use rec = ....env['beesdoo.shift.shift'].new()
rec.field1 = 

This class doesn't seems a good idea. the purpose of the template is to format the data.
If you need some specific formatting don't forget that you can use every method of the object shift. Just define the method that return what you need in a class that inherit beesdoo.shift.shift

remytms: Ok, I will improve that in another PR.

PR #45, opened 2018-06-07
[FIX] label price computation

robinkeunen:
Swiss supermarkets defined tax as price included.
The price computed for the labels (`product.template.total_with_vat`) assumes the vat is price excluded.
Therefore the price on the label counts the tax twice.

### DIFF
--- a/beesdoo_product/models/beesdoo_product.py
+++ b/beesdoo_product/models/beesdoo_product.py
@@ -1,8 +1,9 @@
  # -*- coding: utf-8 -*-
 from openerp import models, fields, api
 from openerp.tools.translate import _
-from openerp.exceptions import UserError
+from openerp.exceptions import UserError, ValidationError
 import uuid
+
 
 class BeesdooProduct(models.Model):
     _inherit = "product.template"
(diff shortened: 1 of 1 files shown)


robinkeunen:
Je n'arrive pas à puller la branche 9.0-fix-label-price-computation sur les serveurs de test de coopiteasy, une idée @remytms @houssine78 ?

```bash
root@ns3311252:~# su odoo
odoo@ns3311252:~/custom/obeesdoo$ git checkout 9.0-fix-label-price-computation
error: pathspec '9.0-fix-label-price-computation' did not match any file(s) known to git.
```

Vu que c'est urgent, je merge en 9.0 pour que Virgine puisse tester.

remytms:
@robinkeunen , try to git fetch before trying to checkout. This may solve the problem.

robinkeunen:
Nope, I tried that but it did not work. I fear we have some access right mixup between `root` and `odoo` on the file system.

remytms:
Ho, yes that kind of sh**. In this case, just run in root `chown -R odoo:odoo /home/odoo/custom`. I will put this command in a cron.

This problem occurs when someone use git as root.

robinkeunen:
This is what I suspected, thank you for the command! 👍

PR #46, opened 2018-09-30
[ADD] custom beesdoo product view

robinkeunen:
I added a new view for beesdoo product because

- the product view is already heavily inherited
- the requested view is customized to Enrico and Bart's specific needs

### DIFF
--- a/beesdoo_product/views/beesdoo_product.xml
+++ b/beesdoo_product/views/beesdoo_product.xml
@@ -72,6 +72,30 @@
         </field>
     </record>
 
+
+    <record model="ir.ui.view" id="beesdoo_product_tree">
+        <field name="name">bees.product.template.tree</field>
+        <field name="model">product.template</field>
+        <field name="priority">14</field>
+        <field name="arch" type="xml">
+            <tree>
+                <field name="name"/>
+                <field name="categ_id"/>
+                <field name="main_seller_id"/>
+                <field name="weight"/>
+                <field name="main_price"/>
+                <field name="main_minimum_qty"/>
+                <field name="uom_id"/>
+                <field name="qty_available"/>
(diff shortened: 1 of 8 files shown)


beesdoo_product/views/beesdoo_product.xml
```diff
@@ -72,6 +72,30 @@
         </field>
     </record>
 
+
+    <record model="ir.ui.view" id="beesdoo_product_tree">
+        <field name="name">bees.product.template.tree</field>
```
tfrancoi: Why don't you inherit from product.product_product_tree_view ?

robinkeunen: Because there are fields I want to remove, added in many different modules. And I want to add more. I figured it was easier and cleaner to start from scratch since this view is not meant to be inherited.

Should I rather move it to a `beesdoo_custom`  module? I heard of people doing that a the Odoo Experience.

houssine78: as it's mainly specific to enrico and bart, it should not be in a generic module. yes put it in a custom module for beesdoo.

tfrancoi: The action: product_normal_action_sell is explicit about the usage of product.product_product_tree_view. This mean if someone open this action, he will not get your specific list view. Since it's for product variant and the bees coop is not using them, I think i doesn't matter much.

robinkeunen: @houssine78 ok, I'll move it to a new module

@tfrancoi Ok thanks for the feedback

robinkeunen:
@tfrancoi @houssine78 Ok with this? I put the model and views in a `beesdoo_custom` module.

PR #47, opened 2018-10-06
[FIX] b_shift: timezone error in shift generation

remytms:
The bug was that some shift are 'misstimezoned' near DST moment and for
timezone far from the UTC time. Now the date and time of a shift is
calculated in the local timezone then converted into UTC time to be
stored in database.

### DIFF
--- a/beesdoo_shift/models/planning.py
+++ b/beesdoo_shift/models/planning.py
@@ -3,9 +3,13 @@
 from openerp import models, fields, api, _
 from openerp.exceptions import UserError
 
-from pytz import UTC
+from pytz import timezone, UTC
+import logging
 import math
 from datetime import datetime, timedelta
+
+_logger = logging.getLogger(__name__)
+
 
 def float_to_time(f):
     decimal, integer = math.modf(f)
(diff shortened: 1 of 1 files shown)


remytms:
I use PEP8, please leave it like this. I didn't touch the rest of the file. :)

But feel free to adapt the code if it's not right for odoo, and feel free to complete the comments if there are not right or inaccurate.

Thanks. :)

tfrancoi:
see https://github.com/beescoop/Obeesdoo/pull/48

remytms:
I think that this PR should be considered in the future as it compute the date and time of a shift more logically and in a more universal way.

For example, if a shift begins at 4:00 AM UTC-5 (like in NYC), this code will do the job properly but the #48 will fail.

Odoo is international, and I think we should code with this in mind. Even if this module is used in Europe for now. :)

tfr-odoo:
I agree with you even if early optimization is the root of evil ;)
Can clean your code and create a new method that take the day, the time and that return the proper time in utc.

remytms:
Good idea. I will do that the next time I work for BEES coop. :)

PR #50, opened 2018-10-09
[FIX] Remove assignation to shift and shift template when you are res…

tfrancoi:
…igning like unsubscribed, you subscribe after resignation you are not resigning anymore ;)


remytms:
@tfrancoi Maybe this PR is already merged or outdated. It should be closed in these cases.

PR #51, opened 2018-10-18
POS2 - afficher si membre peut faire ses courses ou pas

robinkeunen:
ref [POS2 - afficher si membre peut faire ses courses ou pas](https://gestion.coopiteasy.be/web#id=252&view_type=form&model=project.task&action=479)

- [x] extend res.partner backbone model to add state field
- [x] add state to the ActionPadWidget
- [ ] get string value and translate state field (I am getting the technical name)
- [ ] add style to the state div


tfrancoi:
The renaming make the diff less consistent
It would be better to have two commits: one with the renaming of variable and the second that actualy do the job

PR #52, opened 2018-10-19
[CHANGE] -1 instead of -2 when not attempting a shift, -2 only when a…

tfrancoi:
…lert time is not set, remove alert time only once you reach +1


remytms:
@tfrancoi Maybe this PR is already merged or outdated. It should be closed in these cases.

PR #58, opened 2019-02-27
[ADD] pos_require_product_quantity

robinkeunen:
https://gestion.coopiteasy.be/web#id=1196&view_type=form&model=project.task&menu_id=338&action=479

- add require_product_quantity parameter
- sets default order line quantity depending on product category
- pops up when an order line has quantity set to 0

### DIFF
--- a/pos_require_product_quantity/static/src/js/pos.js
+++ b/pos_require_product_quantity/static/src/js/pos.js
@@ -1,0 +1,71 @@
+/*
+  Copyright 2019 Coop IT Easy SCRLfs
+  	    Robin Keunen <robin@coopiteasy.be>
+  License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+*/
+
+odoo.define(
+
+'pos_require_product_quantity.pos_require_product_quantity',
+function (require) {
+"use strict";
+
+var core = require('web.core');
+var models = require('point_of_sale.models');
+var screens = require("point_of_sale.screens");
+
+var _t = core._t;
+var orderline_prototype = models.Orderline.prototype;
+
+models.Orderline = models.Orderline.extend({
+    initialize: function (attr, options) {
+        orderline_prototype.initialize.call(this, attr, options);
+
+        var unit = this.get_unit();
+        if (unit) {
+            if (unit.category_id[1] === 'Unit') {
+                this.set_quantity(1);
+            } else {
+                this.set_quantity(0);
+            }
+        }
+    }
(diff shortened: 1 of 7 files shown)


robinkeunen:
Yo @tfrancoi, @remytms, @houssine78, @PierrickBrun 

Je suis vraiment un gros n00b en js donc faut pas hésiter sur les suggestions. 

Notamment dans le `body` du popup, je ne sais pas comment passer à la ligne pour afficher une liste de produits 😬

pos_require_product_quantity/static/src/js/pos.js
```diff
@@ -0,0 +1,71 @@
...
+var _t = core._t;
+var orderline_prototype = models.Orderline.prototype;
+
+models.Orderline = models.Orderline.extend({
+    initialize: function (attr, options) {
+        orderline_prototype.initialize.call(this, attr, options);
+
+        var unit = this.get_unit();
+        if (unit) {
+            if (unit.category_id[1] === 'Unit') {
```
legalsylvain, line 26: what is unit.category_id[1] ? the name of the category ? if yes, it will not work in non english context ? or did I missed something ?

legalsylvain (commented):
Nice module.

General remark. What about to have a more generic module, named pos_default_quantity. (that is the main objective of this module (setting 1 for Unit and 0 otherwise)). (the require option is more a nice to have feature to avoid dummy lines with null quantity).

We could imagine to make the module more configurable, adding a new float field on product.uom.categ model named pos_default_qty. (set to 1 for Unit categ and to 0 for others, by data noupdate file).

That way, no need to test in the javascript        ``if (unit.category_id[1] === 'Unit') {``
just replace by ``this.set_quantity(categ.pos_default_qty);`` (or something like that).

legalsylvain:
or better, create two modules : 
- pos_require_product quantity that only check if quantity is null. Very generic modules. Can be set in the OCA.
- pos_default_quantity with the ``set_quantity`` options; Can be set under OCA umbrella too.

robinkeunen:
Thanks for the comments Sylvain, here are the two PR:
- [[ADD] pos_require_product_quantity: raise popup if orderline quantity is 0 #328](https://github.com/OCA/pos/pull/328)
- [[ADD] pos_default_quantity: sets default orderline quantity #329](https://github.com/OCA/pos/pull/329)

robinkeunen:
Closing this PR as it was submitted onto OCA/POS.

PR #70, opened 2019-04-16
[FIX] beesdoo_shift: fix first registration shift

nicolasj77:
Lors de la création d'un shift pour un coopérateur qui n'a pas encore de régime de travail, les informations du wizard n'étaient pas conservées. Il fallait repasser par la création d'un shift pour que les informations soient conservées.

### DIFF
--- a/beesdoo_shift/wizard/subscribe.py
+++ b/beesdoo_shift/wizard/subscribe.py
@@ -98,15 +98,15 @@ class Subscribe(models.TransientModel):
             self.cooperator_id.sudo().write({'subscribed_shift_ids' : [(6,0, [self.shift_id.id])]})
         if self.working_mode != 'regular':
             #Remove existing shift then subscribe to the new shift
-            self.cooperator_id.sudo().write({'subscribed_shift_ids' : [(5,)]})
+            self.cooperator_id.sudo().write({'subscribed_shift_ids': [(5,)]})
             
 
         data = {
-            'info_session' : self.info_session,
+            'info_session': self.info_session,
             'info_session_date': self.info_session_date,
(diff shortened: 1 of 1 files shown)


houssine78 (commented):
👍 for the change made by @nicolasj77 but I need @remytms to check the PR for the code change in the controller to see whether is accurate or not...

houssine78 (approved):

beesdoo_shift/wizard/subscribe.py
```diff
@@ -121,10 +121,16 @@ def subscribe(self):
...
+            [('cooperator_id', '=', self.cooperator_id.id)])
         if status_id:
             status_id.sudo().write(data)
         else:
-            self.env['cooperative.status'].sudo().create(data)
-            
-
+            status_id = coop_status_obj.sudo().create(data)
+            # Normally the write method is not necessary here.
+            # But it does not work without it. You have to make 2 registration
```
tfrancoi, line 132: Maybe it would be better to investigate the root reason why this does not work instead of writing a piece of code that doesn't seems to make sense

nicolasj77: We did investigate but it took a lot of time. We did prefer to do it like this to not exceed the estimate time. I can maybe review it again with Rémy when he comes back.

houssine78: Yes in fact, we needed to fix it and the dedicated budget time to investigate and fix was consumed. So we chosen to fix it like this and to put a todo to investigate it later during a benevolent shift.

If you feel to investigate this Thibault, we'll be so greatful.

PR #76, opened 2019-05-14
[FIX] b_w_shift: Issue with settings

remytms:
There was an issue that shows an error message when trying to modify the settings for shift in the website configuration.

This is due to a missing sequence in the database. I don't know why this sequence where not present. But a simple query create it if it doesn't exists.


remytms:
@nicolasj77 Je merge pour avancer, mais jette un œil à la solution. Je pense pas que ça vienne d'un problème de ton code.

PR #84, opened 2019-05-31
[ADD] b_easy_my_coop: Info Session in new cooperator form

remytms:
This adds, in the new cooperator form in the website, a checkbox that
the user can check if he already has taken part to an information
session.

This information will be saved in the partner object and will be used to
automatically check the info_session field of the cooperator_status when
it will be created by the wizard.


remytms:
This can only work with this PR: https://github.com/coopiteasy/vertical-cooperative/pull/2

PR #100, opened 2019-09-12
[FIX] beesdoo_shift : bug occuring when trying to save a temporary exempt

elouan-lb:
Caused by a missing sequence in database, fixed with post-migrate file.


remytms (approved):
Yeah !

PR #102, opened 2019-09-22
[fix]: working dockerfile and docker-compose

aborsu:
Refactored the docker file and the docker-compose and added documentation.
It's working but I'm still having some issues installing all the apps.

I was thinking of dropping by next week for an hour again to try and fix the last bits.


aborsu:
@remytms I would appreciate any feedback on this as It took me a loooong time to get it to work.
I'm pretty happy with the results but don't hesitate to tell me if anything can be improved

aborsu:
@remytms @vdewulf I'm still missing one dependency '10n_be_vat_reports'
Do you know where that comes from? A quick google and github search didn't help.
Also lots of unique, not null constraints and missing indices warnings when starting.
Is that normal?

PR #91, opened 2019-06-08
[ADD] beesdoo_shift: warn when assigning a second shift

aborsu:
Warn when adding a shift to a worker who is already subscribed to one.

PR #104, opened 2019-10-03
fix: ghost error message

aborsu:
The error message was displayed because the invisible condition is only
triggered when all the fields have been rendered and the javascript is
evaluated.

There is nothing I can do about that. Instead what I did was remove the
css box around the text so that if no text is present nothing is shown.
Added the error message as a field so that it only appears when the
attrs are evaluated.

If I leave the alert class an empty box is shown instead of the full message.
If you pay attention, at the same time than the error message is shown, so are the buttons and the or in between. This is inherent to how odoo templates are displayed.

### DIFF
--- a/beesdoo_shift/__openerp__.py
+++ b/beesdoo_shift/__openerp__.py
@@ -13,7 +13,7 @@
     'website': "https://github.com/beescoop/Obeesdoo",
 
     'category': 'Cooperative management',
-    'version': '9.0.1.2.3',
+    'version': '9.0.1.2.4',
 
     'depends': ['beesdoo_base'],
 
--- a/beesdoo_shift/wizard/subscribe.py
+++ b/beesdoo_shift/wizard/subscribe.py
@@ -48,7 +48,11 @@ class Subscribe(models.TransientModel):
         return
 
     def _get_nb_shifts(self):
-        return len(self.env['res.partner'].browse(self._context.get('active_id')).subscribed_shift_ids)
+        if len(self.env['res.partner'].browse(self._context.get('active_id')).subscribed_shift_ids) > 1:
+            return _("Current worker has more than one shift, "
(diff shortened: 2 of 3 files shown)


aborsu:
fixes #91 @vdewulf

aborsu:
image

The new error message is slightly more discreet.

vdewulf:
@aborsu Thanks for your help and explanation!
@elouan-lb please tell me when I can test it on BEES coop test servor.

elouan-lb (approved):
As far as I know Odoo's behaviour, this workaround seems great. Thank you !

elouan-lb (dismissed):

elouan-lb:
@vdewulf it's added on the test servor. We could merge the fixup after your test !

vdewulf:
Tested on bees-coop test servor.
It's not working, no warning message is displayed when I want to attribute a new shift template to a regular worker that is already subscribed to a shift template.
I talked with @elouan-lb and he seems he sees the issue.
Thanks anyway for having checked this issue.
The ghost does not appear anymore :)

elouan-lb:
@aborsu the problem may come from the non-update of the warning field. You could have a look at computed fields : https://www.odoo.com/documentation/online/reference/orm.html#computed-fields

remytms:
@elouan-lb Thank you for reviewing this. I think this should be merged. Did you integrate this changes when porting `beesdoo_shift` to 12.0 ? If not it would be great to port them in the migration branch of `beesdoo_shift` after merging it into 9.0.

elouan-lb (commented):
According to @vdewulf the fix-up is not working, would you mind having a look at this ? 

Then I will integrate it to https://github.com/beescoop/Obeesdoo/tree/12.0-mig-beesdoo_shift as @remytms wisely suggested.

PR #119, opened 2020-01-02
[FIX] b_shift : next_countdown computation

elouan-lb:
When `next_countdown` is the same day than the current one, it should be the one displayed (and not the one 28 days later).


remytms (approved):
Thanks. :)

PR #122, opened 2020-01-06
[FIX] b_shift : automatic unsubscribe from next shifts 

elouan-lb:
When a worker is changing to _unsubscribed_ status. Doesn't unsubscribe him from present shift anymore, only from shifts starting right after unsubscription call.

Future shifts status cannot be modified anymore (only for _confirmed_ / _cancelled_). Thus we don't need to reset `revert_info` for unsubscribed shifts as it is never set.

### DIFF
--- a/beesdoo_shift/models/task.py
+++ b/beesdoo_shift/models/task.py
@@ -62,6 +62,16 @@ class Task(models.Model):
                 "Compensation Shift."
             )
 
+    @api.constrains("state")
+    def _lock_future_task(self):
+        start_time_dt = fields.Datetime.from_string(self.start_time)
+        if datetime.now() < start_time_dt:
+            if self.state in ["done", "absent_2", "absent_1", "absent_0"]:
+                raise UserError(_(
+                    "You cannot set shift state to 'present' "
+                    "or 'absent' for a future shift."
+                ))
+
     @api.constrains('is_regular', 'is_compensation')
     def _check_compensation(self):
         for task in self:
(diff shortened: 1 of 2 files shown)


beesdoo_shift/models/task.py
```diff
@@ -104,19 +117,27 @@ def _read_group_stage_id(self, ids, domain, read_group_order=None, access_rights
...
-        # date_domain = [('worker_id', 'in', worker_ids), ('start_time', '>=', today)]
-        date_domain = [('start_time', '>=', today)]
+    def unsubscribe_from_today(self, worker_ids, today=None, end_date=None, now=None):
+        """
+            If 'now' is set (must be a datetime), start_time and end_time
+            are replaced and worker is unsuscribed from 'now'.
+        """
+        if not now:
+            today = today or fields.Datetime.now()
+            today += ' 00:00:00' # Only if today not None
```
remytms: Here, it should be:
```python
today = today or fields.Date.today()
today += ' 00:00:00'
```
Because you can not add `' 00:00:00'` to a `datetime`.

beesdoo_shift/models/task.py
```diff
@@ -104,19 +117,27 @@ def _read_group_stage_id(self, ids, domain, read_group_order=None, access_rights
...
-        today = today or fields.Date.today()
-        today += ' 00:00:00'
-        if end_date:
-            end_date += ' 23:59:59'
-        # date_domain = [('worker_id', 'in', worker_ids), ('start_time', '>=', today)]
-        date_domain = [('start_time', '>=', today)]
+    def unsubscribe_from_today(self, worker_ids, today=None, end_date=None, now=None):
+        """
+            If 'now' is set (must be a datetime), start_time and end_time
+            are replaced and worker is unsuscribed from 'now'.
```
remytms:
```python
"""
Unsubscribe workers from *worker_ids* from all shift that start *today* and later.
If *end_date* is given, unsubscribe workers from shift between *today* and *end_date*.
If *now* is given workers are unsubscribed from all shift that start *now* and later.
If *now* is given, *end_date* is not taken into account.

:type today: fields.Date
:type end_date: fields.Date
:type now: fields.Datetime
"""
```

Something like that will be clearer.

beesdoo_shift/models/task.py
```diff
@@ -104,19 +117,27 @@ def _read_group_stage_id(self, ids, domain, read_group_order=None, access_rights
...
+            today += ' 00:00:00' # Only if today not None
+            if end_date:
+                end_date += ' 23:59:59'
+        else:
+            if len(now) != 19:
+                raise UserError (_("'Now' must be a datetime."))
+            today = now
+        date_domain = [('start_time', '>', today)]
         if end_date:
             date_domain.append(('end_time', '<=', end_date))
```
remytms: Here you should not use `end_date` if `now` is given, or you should ensure that `end_date` is a datetime format like you do when `now` is not given.

remytms (changes requested):
Checks my comment. :)

elouan-lb:
> Checks my comment. :)

Thank you for the review, changes were made !

remytms (approved):

PR #134, opened 2020-01-27
[12.0] [MIG] beesdoo_easy_my_coop

StCyr:
**Do not try merging yet.**

Following questions/problems still need to be answered/solved:

1. Migration 11.0->12.0: doc says `All <label> elements in views must have a for="" attribute.`. I'm wondering if a text is mandatory between the quotes in `for=""` (concerns file `views/subscription_templates.xml`)
2. Module won't load. I think it's because it depends on module `beesdoo_shift` which is still in version 9.0 in branch `12.0-mig-obeesdoo`


beesdoo_easy_my_coop/views/partner.xml
```diff
@@ -0,0 +1,47 @@
...
+
+    <record model="ir.ui.view" id="beesdoo_partner_form_view_remove_duplicate_header">
+        <field name="name">beesdoo.partner.form.view.remove_duplicate_header</field>
+        <field name="model">res.partner</field>
+        <field name="inherit_id" ref="beesdoo_shift.super_coop_partner_inherited_view_form" />
+        <field name="arch" type="xml">
+            <xpath expr="header" position="replace" />
+            <xpath expr="//field[@name='state']" position="before">
+                <button string="Create Subscription" name="%(easy_my_coop.action_view_create_subscription)d"
+                                type="action" context="{'default_active_id': active_id}" groups="easy_my_coop.group_energiris_manager"/>
```
houssine78: groups name changed in 12. see https://github.com/coopiteasy/vertical-cooperative/blob/12.0/easy_my_coop/security/res_groups.xml here.

beesdoo_easy_my_coop/views/partner.xml
```diff
@@ -0,0 +1,47 @@
...
+                 <button string="Update info" name="%(easy_my_coop.action_view_update_partner_info)d"
+                                type="action" context="{'default_active_id': active_id}" groups="easy_my_coop.group_energiris_manager"/>
```
houssine78: same here

beesdoo_easy_my_coop/views/partner.xml
```diff
@@ -0,0 +1,47 @@
...
+        <field name="arch" type="xml">
+            <filter name="type_a" position="replace">
+                <filter name="type_a" string="Cooperator type A" domain="[('cooperator_type', '=', 'share_a'), ('member', '=', True)]" />
+            </filter>
+            <filter name="type_b" position="replace">
+                <filter name="type_b" string="Cooperator type B" domain="[('cooperator_type', '=', 'share_b'), ('member', '=', True)]" />
+            </filter>
+            <filter name="type_c" position="replace">
+                <filter name="type_c" string="Cooperator type C" domain="[('cooperator_type', '=', 'share_c'), ('member', '=', True)]" />
+            </filter>
```
houssine78: it's no more hard coded an it's not generic. so this search view can be removed.

beesdoo_easy_my_coop/views/partner.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+	<record model="ir.ui.view" id="beesdoo_partner_form_view_remove_duplicate_cooperator_type">
+		<field name="name">beesdoo.partner.form.view.remove_duplicate_cooperator_type</field>
+		<field name="model">res.partner</field>
+		<field name="inherit_id" ref="beesdoo_base.beesdoo_partner_form_view" />
+		<field name="arch" type="xml">
+			<field name="cooperator_type" position="replace" />
+		</field>
```
houssine78: I would rather see if it still a duplicate in the view. Anyway it's no more a problem on odoo 12.

beesdoo_easy_my_coop/models/res_partner.py
```diff
@@ -0,0 +1,44 @@
...
+        share_type_list = [('','')]
+        for share_type in self.env['product.product'].search([('is_share','=',True)]):
+            share_type_list.append((share_type.default_code, share_type.short_name))
+        return share_type_list
+
+    @api.noguess
+    def _auto_init(self, cr, context=None):
+        cr.execute("ALTER TABLE res_partner DROP COLUMN IF EXISTS cooperator_type")
+        res = super(Partner, self)._auto_init(cr, context=context)
+        return res
```
houssine78: can be removed as code is pretty similar now in easy my coop

beesdoo_easy_my_coop/models/res_partner.py
```diff
@@ -0,0 +1,44 @@
+from odoo import models, fields, api, _
+
+class Partner(models.Model):
+    _inherit = 'res.partner'
+
+    cooperator_type = fields.Selection(selection='_get_share_type', compute='_compute_share_type', string='Cooperator Type', store=True)
```
houssine78: remove field override

beesdoo_easy_my_coop/models/res_partner.py
```diff
@@ -0,0 +1,44 @@
...
+    can_shop = fields.Boolean(compute='_can_shop', store=True)
+    info_session_confirmed = fields.Boolean(
+        string="Confirmed presence to info session",
+        default=False,
+    )
+
+    @api.depends('cooperator_type', 'cooperative_status_ids', 'cooperative_status_ids.can_shop')
+    def _can_shop(self):
+        for rec in self:
+            if rec.cooperator_type == 'share_b':
```
houssine78: this should not be hardcoded anymore. a boolean can_shop on the share(product.template) should do the work

houssine78 (changes requested):
some adaptation should be brough. Thanks for the work so far !

houssine78 (approved):

houssine78 (approved):

houssine78 (approved):

beesdoo_easy_my_coop/models/res_partner.py
remytms: There is already a flag in easy_my_coop share product named `customer` that do the same thing then `can_shop`. So I've removed `can_shop` on the share product.

See : https://github.com/coopiteasy/vertical-cooperative/blob/12.0/easy_my_coop/models/product.py#L21

houssine78: as far that the can_shop on res_partner is handled and correctly mapped to customer field from the product, it is fine for me.

houssine78 (approved):

PR #146, opened 2020-04-17
[12.0] [MIG] beesdoo_pos

mclaeysb:
I removed (together with their security and translation files)

- 'Bill Value' (standard in v12)
- 'Send Orders by email' (replace with OCA module)

Functionallity that's still in place:

- Displaying eaters on POS ActionpadWidget and PaymentScreenWidget (altought not functional anymore - @tfr-odoo or @vvrossem, could you take a look at this? My initial debugging didn't go further then checking that the spans's for this information`<span class="customer-delegate1"></span>` are in place and the `get_eater()` and `_get_eater()` function work, but somehow they are not called.)
- Some modification to `default_barcode_pattern.xml` encoding.


vvrossem:
@mclaeysb 
After installing `beesdoo_pos` and opening a POS session, the browser console shows a warning about `web.DataModel`:
beesdoo_pos
It's deprecated and you need to use `web.rpc` instead:
- [Missing dependencies web.DataModel in odoo11](https://www.odoo.com/forum/help-1/question/missing-dependencies-web-datamodel-in-odoo11-133658) 
- [web.DataModel is missing odoo11](https://stackoverflow.com/questions/50403242/web-datamodel-is-missing-odoo11)

`beesdoo.pos` should load smoothly after the modification 🕺

remytms:
@mclaeysb I think that @vvrossem is a better reviewer for this task.

mclaeysb:
Fixed, thanks @vvrossem for the help 🎉 !

beesdoo_pos/static/src/js/beesdoo.js
```diff
@@ -1,7 +1,7 @@
 odoo.define('beescoop.pos', function (require) {
```
vvrossem: OCA convention is : module_name.type (type = screens / models / etc...)

In our case : `odoo.define('beesdoo_pos.screens', function (require) {...})`

beesdoo_pos/static/src/js/beesdoo.js
```diff
@@ -1,7 +1,7 @@
...
     "use strict";
     var module   = require("point_of_sale.screens");
```
vvrossem: rename `module` occurrences in `screens`
`var screens = require("point_of_sale.screens");`

beesdoo_pos/static/src/js/beesdoo.js
```diff
@@ -1,7 +1,7 @@
...
-    var Model = require('web.DataModel');
+    var rpc = require('web.rpc')
```
vvrossem: I am unsure about this, but since we use `this._rpc`, `rpc` import can be removed

vvrossem (changes requested):
LGTM. Just some renaming to comply with OCA conventions

PR #147, opened 2020-04-17
[12.0][MIG] removing pos_price_to_weight in favour of OCA

mclaeysb:
Note that there are extra commits by Houssine on `beescoop/Obeesdoo/pos_price_to_weight` in v9, making sure that the price with vat is used. The aim is to propose those modification to the OCA in a PR.


remytms (approved):
LGTM. :)

remytms:
See #148

PR #148, opened 2020-04-29
[REM] pos_price_to_weight: remove in favour of OCA version

mclaeysb:
Removing pos_price_to_weight, will use the same module but from OCA, and adapt it to use `product.total_with_vat` rather then `product.price`


remytms (approved):

remytms:
See also https://github.com/OCA/pos/pull/485

PR #151, opened 2020-05-13
[12.0] [MIG] beesdoo_crelan_csv

mclaeysb:
Adds ability to import Crélan csv.

Please review the part where I set `self.init_balance` and strip the currency from the string.

Must be tested on a database where localisation has EUR.


remytms (approved):
LGTM

PR #152, opened 2020-05-21
[FIX] remove style in separator tag.

houssine78:
The first one break the ability to click on the button. And style seems
to have no effect on v12.

[task](https://gestion.coopiteasy.be/web#id=3971&view_type=form&model=project.task)

### DIFF
--- a/beesdoo_shift_attendance/__manifest__.py
+++ b/beesdoo_shift_attendance/__manifest__.py
@@ -18,7 +18,7 @@
     'website': "https://github.com/beescoop/Obeesdoo",
 
     'category': 'Cooperative management',
-    'version': '12.0.1.0.0',
+    'version': '12.0.1.0.1',
 
     'depends': [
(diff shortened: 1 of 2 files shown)


mclaeysb (commented):
Button LGTM 
I can, however, find some examples of `style` in v12 (e.g. in `coopiteasy/obeesdoo/beesdoo_shift_attendance/views/attendance_sheet.xml`) so I'm not sure it has no effect. But that's not the crucial part here I guess...

houssine78:
> 
> 
> Button LGTM
> I can, however, find some examples of `style` in v12 (e.g. in `coopiteasy/obeesdoo/beesdoo_shift_attendance/views/attendance_sheet.xml`) so I'm not sure it has no effect. But that's not the crucial part here I guess...

you should point me to the line but anyway we'll look at this in a specific task later on.

remytms (approved):
LGTM

PR #157, opened 2020-06-10
[12.0][FIX] b_shift, b_emc, macavrac_base: is_worker computed field

vvrossem:
### [Task](https://gestion.coopiteasy.be/web#id=4706&view_type=form&model=project.task)

This PR brings changes on `res.partner` in the following modules:
- `beesdoo_shift`
- `beesdoo_easy_my_coop`
- `macavrac_base`

#### In `beesdoo_shift`:
- `is_worker`, initially `related`, is now `compute="_compute_is_worker"` (according to `worker_store` value) and `store=True` in order to not implement any `search=`. It his hidden in the view.
- `worker_store` is added to the view

#### In `beesdoo_easy_my_coop`:
- `_compute_is_worker()` overrides the parent behavior and `search="_search_worker"` is removed
- `worker_store` is hidden in the view
- `is_worker` is visible

#### In `macavrac_base`:
- `search="_search_is_worker"` is removed 

#### Context:
We had to implement those changes because, in `beesdoo_easy_my_coop`, `worker_store` remained `False` despite being assigned to the same value as `is_worker` in `_compute_is_worker()`. For a reason we couldn't explain, once the method was executed, the value was assigned back to `False`.

This was problematic since the `_search_worker()` was done on the `worker_store`. When a worker was subscribing to a shift, he was not appearing in it:
beesdoo_easy_my_coop-error

### DIFF
--- a/beesdoo_easy_my_coop/models/res_partner.py
+++ b/beesdoo_easy_my_coop/models/res_partner.py
@@ -12,12 +12,6 @@ class Partner(models.Model):
         string="Confirmed presence to info session",
         default=False,
     )
-    is_worker = fields.Boolean(
-        compute="_is_worker",
-        search="_search_worker",
-        readonly=True,
-        related=""
-    )
 
     def _cooperator_share_type(self):
         """
@@ -33,12 +27,9 @@ class Partner(models.Model):
         return share_type
 
     @api.depends(
-        'share_ids',
-        'share_ids.share_product_id',
-        'share_ids.share_product_id.default_code',
-        'share_ids.share_number',
+        "cooperator_type"
     )
-    def _is_worker(self):
+    def _compute_is_worker(self):
         """
         Return True if the partner can participate tho the shift system.
         This is defined on the share type.
@@ -47,13 +38,8 @@ class Partner(models.Model):
             share_type = rec._cooperator_share_type()
             if share_type:
                 rec.is_worker = share_type.allow_working
-                rec.worker_store = share_type.allow_working
             else:
                 rec.is_worker = False
-                rec.worker_store = False
-
-    def _search_worker(self, operator, value):
-        return [('worker_store', operator, value)]
 
     @api.depends(
         "cooperative_status_ids",
--- a/beesdoo_shift/models/res_partner.py
+++ b/beesdoo_shift/models/res_partner.py
@@ -13,7 +13,7 @@ class ResPartner(models.Model):
     _inherit = 'res.partner'
 
     worker_store = fields.Boolean(default=False)
-    is_worker = fields.Boolean(related="worker_store", string="Worker", readonly=False)
+    is_worker = fields.Boolean(string="Worker", compute="_compute_is_worker", store=True)
     can_shop = fields.Boolean(string="Is worker allowed to shop?", compute="_compute_can_shop", store=True)
     cooperative_status_ids = fields.One2many('cooperative.status', 'cooperator_id', readonly=True)
     super = fields.Boolean(related='cooperative_status_ids.super', string="Super Cooperative", readonly=True, store=True)
@@ -24,6 +24,11 @@ class ResPartner(models.Model):
     state = fields.Selection(related='cooperative_status_ids.status', readonly=True, store=True)
     extension_start_time = fields.Date(related='cooperative_status_ids.extension_start_time', string="Extension Start Day", readonly=True, store=True)
     subscribed_shift_ids = fields.Many2many('beesdoo.shift.template')
+
+    @api.depends("worker_store")
+    def _compute_is_worker(self):
+        for rec in self:
+            rec.is_worker = rec.worker_store
 
     @api.depends("cooperative_status_ids")
     def _compute_can_shop(self):
--- a/beesdoo_easy_my_coop/__manifest__.py
+++ b/beesdoo_easy_my_coop/__manifest__.py
@@ -26,7 +26,8 @@
         'views/res_company.xml',
         'views/subscription_request.xml',
         'views/subscription_templates.xml',
-        'views/product.xml'
+        'views/product.xml',
+        'views/res_partner.xml'
     ],
     'demo': [
         'demo/product_share.xml',
--- a/beesdoo_easy_my_coop/tests/test_res_partner.py
+++ b/beesdoo_easy_my_coop/tests/test_res_partner.py
@@ -143,7 +143,7 @@ class TestResPartner(TransactionCase):
             "beesdoo_base.res_partner_cooperator_1_demo"
         )
         # Run computed field
-        coop1._is_worker()
+        coop1._compute_is_worker()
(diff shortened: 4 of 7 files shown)


beesdoo_shift/models/res_partner.py
```diff
@@ -13,7 +13,7 @@ class ResPartner(models.Model):
     _inherit = 'res.partner'
 
     worker_store = fields.Boolean(default=False)
-    is_worker = fields.Boolean(related="worker_store", string="Worker", readonly=False)
+    is_worker = fields.Boolean(computed="_compute_is_worker", string="Worker", readonly=True, store=True)
```
houssine78: I think that we can remove readonly attribute as it True by default for computed fields

beesdoo_easy_my_coop/models/res_partner.py
```diff
@@ -38,7 +32,7 @@ def _cooperator_share_type(self):
         'share_ids.share_product_id.default_code',
         'share_ids.share_number',
```
houssine78: simpler to set the depedency on cooperator_type fields as this is what interested us. And that computation of cooperator_type is triggered on those fields.

beesdoo_easy_my_coop/models/res_partner.py
```diff
@@ -47,13 +38,8 @@ def _is_worker(self):
             share_type = rec._cooperator_share_type()
             if share_type:
                 rec.is_worker = share_type.allow_working
-                rec.worker_store = share_type.allow_working
             else:
                 rec.is_worker = False
-                rec.worker_store = False
```
vvrossem, line 53: Shoud I leave `worker_store` for coherency ?

houssine78: nope not needed anymore !

beesdoo_shift/models/res_partner.py
houssine78: you don't want to remove readonly=True ?

houssine78 (approved):

beesdoo_shift/models/res_partner.py
vvrossem: it has been removed with the force-push

vvrossem:
Hi @tfrancoi, could you have a look on this PR? Thank you :bowing_man:

tfrancoi:
> Hi @tfrancoi, could you have a look on this PR? Thank you bowing_man

Yes I will look at it. 
Can you describe in details the scenario to reproduce the issue because it don't think this is the proper way to fix this. Before we have to understand the root cause of the issue. 

worker_store should not be use by the beescoop, this field is only use in case of beesdoo_worker_status is not installed.

remytms:
@tfrancoi In the current 12.0 branch (so, without this refactoring). I implemented `is_worker` in the `beesdoo_easy_my_coop` module. This module computes the `is_worker` property, looking to the configuration done on the share that the user owns. See: https://github.com/beescoop/Obeesdoo/blob/12.0/beesdoo_easy_my_coop/models/res_partner.py

To problem occurs in the `_search_worker()` method. What I understand is that the `worker_store` field purpose is to write the `is_worker` state in the database (like a `store=True` in on the `is_worker` field). With this field (`worker_store`) I could easily write the `_search_worker()` method as you can see in the 12.0 branch.

The issue is that `worker_store` seams not to be updated by the `_is_worker()` method, but the `is_worker` field is correctly updated. Therefore, search (or filter) on the `is_worker` field does not work.

I wrote test, to check this behaviour, here: https://github.com/beescoop/Obeesdoo/blob/12.0/beesdoo_easy_my_coop/tests/test_res_partner.py#L160-L179
This tests succeed, but may be because `is_worker` field is properly set in the demo data: https://github.com/beescoop/Obeesdoo/blob/12.0/beesdoo_shift/demo/workers.xml
I'm not sure about the computed fields and the demo data when running tests.

By the way, this demo data file has issue because it's based on data from `beesdoo_base`. It will be fixed later on. So to test the `beesdoo_easy_my_coop` module, first install the `beesdoo_base` module, then the `beesdoo_easy_my_coop` (that will trigger the installation of `beesdoo_shift`).

The issue appears more clearly when looking to a task template. The field `worker_ids` will be always empty: https://github.com/beescoop/Obeesdoo/blob/12.0/beesdoo_shift/models/planning.py#L92
This is because the domain on this field return always an empty set because of the issue above.

Hope this clarify the issue. Any proposal are welcome, even if we found our solution not too bad.

Thanks in advance for you review and hints. :)

tfrancoi:
@remytms worker_store shouldn't be part of the problem. 
this field is only use when no other module then beesdoo_shift is install to let the user manually set the value and it's only for the internal implementation in that case

If you redefine is_worker in beesdoo_easy_my_coop it should be fine.
your compute field should not care about the worker_store, should be store. If it's store maybe you have to define search=None and if it's not store it should of course define your own search method.

houssine78:
@tfrancoi I think that the purpose of setting it like you did was for us. Anyway I don't think that the modification has impact on your code and it simplify a bit the code. No need to define it related at the first place and override it to make it computed from both side, macavrac and beescoop. making it computed from the begining and make the computed is_worker field on the worker_store if the module is used in standalone is not a big deal.

Ok maybe we can avoid setting it store=True if you prefere not to. But I think that setting it related at the first place is useless from both of us as we both override it to set related=''.

tfrancoi:
I don't know if it's a good idea or not to make it store. Since it's a boolean, the search fonction is not that hard to implement and we probably not really interested to group by that field. So I didn't see the point to make it store

No it's not a big deal to make it compute instead of related but in that case the inverse function should be defined (instead of showing worker_store which is internal implementation)

houssine78:
@tfrancoi make it store is only to have it in the database. As you know some external tools (jasper) tackle the bees database and may need it.

Ok thanks we'll take your comment into account and adapt the PR accordingly !

kisses bad boy !

tfrancoi:
Those tools cannot check directly the source of the compute field ? the share of cooperator which is as easy to get. 

But I'm not against storing the field.

houssine78:
@tfrancoi they can but they don't want to know or rely on it. It more queries to get it while they could have it directly on res partner. We don't set those tools.

tfrancoi:
@remytms Thx for the scenario I manage to reproduce the issue.
The issue is only in beesdoo_easy_my_coop module.

You define is_worker based on the share line and the product allow_working fine
this the definition of beescoop of a worker.

Then you need to change the search function to match that behavior

find all the partner that have a share.line with a product allow_working.

worker_store is a field that you don't have to care about at all if you redefine is_worker
I don't see why you would use it in the search function

You can see my suggestion here https://github.com/beescoop/Obeesdoo/pull/159

You don't need to touch beesdoo_shift or macavrac_base.

Of course if you feel that you need to make is_worker store, you can do it in beesdoo_easy_my_coop, just be carefull with the depends.

PR #159, opened 2020-06-13
[FIX] beesdoo_easy_my_coop: define properly search is_worker

tfrancoi:
Define properly the search function for is_worker
Find the share line with a product allow working
and then find all the partner link to those lines


remytms:
@tfrancoi Thanks. Will be merged after testing.

PR #164, opened 2020-07-31
[12.0][ADD] beesdoo_product: profit margin on suppliers

mclaeysb:
Add a 'Suggested exVAT Price' field on products, and a 'Product Margin' field on Partners (Vendors) and Product Categories. The first margin is used if set, otherwise the second margin (which has a default value) is used.

### DIFF
--- a/beesdoo_product/__manifest__.py
+++ b/beesdoo_product/__manifest__.py
@@ -2,6 +2,7 @@
 #   - Elouan Lebars <elouan@coopiteasy.be>
 #   - Rémy Taymans <remy@coopiteasy.be>
 #   - Houssine BAKKALI <houssine@coopiteasy.be>
+#   - Manuel Claeys Bouuaert <manuel@coopiteasy.be>
 #   - Elise Dupont
 #   - Thibault François
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
@@ -13,7 +14,7 @@
     "author": "Beescoop - Cellule IT, Coop IT Easy SCRLfs",
     "website": "https://github.com/beescoop/Obeesdoo",
     "category": "Sales",
-    "version": "12.0.1.0.0",
+    "version": "12.0.1.0.1",
     "depends": ["beesdoo_base", "product", "sale", "point_of_sale"],
     "data": [
         "data/product_label.xml",
--- a/beesdoo_product/__manifest__.py
+++ b/beesdoo_product/__manifest__.py
@@ -21,6 +21,7 @@
         "data/product_sequence.xml",
         "views/beesdoo_product.xml",
         "views/assets.xml",
+        "views/res_config_settings.xml",
         "wizard/views/label_printing_utils.xml",
         "security/ir.model.access.csv",
     ],
--- a/beesdoo_product/models/__init__.py
+++ b/beesdoo_product/models/__init__.py
@@ -1,1 +1,2 @@
 from . import beesdoo_product
+from . import res_conig_settings
--- a/beesdoo_product/models/__init__.py
+++ b/beesdoo_product/models/__init__.py
@@ -1,2 +1,2 @@
 from . import beesdoo_product
-from . import res_conig_settings
+from . import res_config_settings
--- a/beesdoo_product/models/beesdoo_product.py
+++ b/beesdoo_product/models/beesdoo_product.py
@@ -11,6 +11,19 @@ from odoo.tools.translate import _
 _logger = logging.getLogger(__name__)
 
 
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    profit_margin = fields.Float(string="Product Margin [%]")
+
+    @api.multi
+    @api.constrains("profit_margin")
+    def _check_margin(self):
+        for product in self:
+            if product.profit_margin < 0.0:
+                raise UserError(_("Percentages for Profit Margin must > 0."))
+
+
 class BeesdooProduct(models.Model):
     _inherit = "product.template"
 
@@ -252,12 +265,17 @@ class BeesdooProduct(models.Model):
         for product in self:
             suppliers = product._get_main_supplier_info()
             if len(suppliers) > 0:
-                product.suggested_price = (
-                    suppliers[0].price * product.uom_po_id.factor
-                ) * (
-                    1
-                    + suppliers[0].product_tmpl_id.categ_id.profit_margin / 100
+                price = suppliers[0].price
+                profit_margin_supplier = suppliers[0].name.profit_margin
+                profit_margin_product_category = suppliers[
+                    0
+                ].product_tmpl_id.categ_id.profit_margin
+                profit_margin = (
+                    profit_margin_supplier or profit_margin_product_category
                 )
+                product.suggested_price = (
+                    price * product.uom_po_id.factor
+                ) * (1 + profit_margin / 100)
 
 
 class BeesdooScaleCategory(models.Model):
--- a/beesdoo_product/models/beesdoo_product.py
+++ b/beesdoo_product/models/beesdoo_product.py
@@ -72,10 +72,8 @@ class BeesdooProduct(models.Model):
 
     note = fields.Text("Comments")
 
-    # S0023 : List_price = Price HTVA, so add a suggested price
-    list_price = fields.Float(string="exVAT Price")
     suggested_price = fields.Float(
-        string="Suggested exVAT Price", compute="_compute_cost", readOnly=True,
+        string="Suggested Price", compute="_compute_cost", readOnly=True,
         help="""
         This field computes a suggested price based on the 'Product Margin' 
         field on Partners (Vendors), if it's set, or otherwise on the 'Product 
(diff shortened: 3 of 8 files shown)


beesdoo_product/models/beesdoo_product.py
```diff
@@ -11,6 +11,19 @@
...
+    _inherit = "res.partner"
+
+    profit_margin = fields.Float(string="Product Margin [%]")
+
+    @api.multi
+    @api.constrains("profit_margin")
+    def _check_margin(self):
+        for product in self:
+            if product.profit_margin < 0.0:
+                raise UserError(_("Percentages for Profit Margin must > 0."))
```
houssine78: From my understanding error message should be "Percentages for Profit Margin must >= 0."

beesdoo_product/models/beesdoo_product.py
```diff
@@ -252,12 +265,17 @@ def _compute_cost(self):
...
+                profit_margin_supplier = suppliers[0].name.profit_margin
+                profit_margin_product_category = suppliers[
+                    0
+                ].product_tmpl_id.categ_id.profit_margin
+                profit_margin = (
+                    profit_margin_supplier or profit_margin_product_category
                 )
+                product.suggested_price = (
+                    price * product.uom_po_id.factor
+                ) * (1 + profit_margin / 100)
```
houssine78: I know it was there but why 1 + profit_margin. having 20 of profit margin make it be of 21. So I don't get it...

from my understanding it should be 1 * profit_margin...

houssine78 (commented):
question and remarks

houssine78:
@mclaeysb could you have a quick look at my question and remark ? thanks ! :)

beesdoo_product/models/beesdoo_product.py
mclaeysb: That's right! I'll fix it.

beesdoo_product/models/beesdoo_product.py
mclaeysb: Order of computations: first the devision, then the addition!
`profit_margin` = 20 => `(1 + profit_margin / 100)` = 1 + 20/100 = 1.20 ✔️

beesdoo_product/models/beesdoo_product.py
mclaeysb: It was a copy pase from line 324. I'll fix it there as well.

beesdoo_product/models/beesdoo_product.py
houssine78: you're right !

mclaeysb:
@houssine78 FYI the last commit: we removed the explicit mention of exVAT from the beesdoo_product module, since it depends on the general settings.

See also: https://gestion.coopiteasy.be/web?token=R0Qs92cjQIn6BsLb5XTZ&db=coopiteasy#id=1854&view_type=form&model=project.task&menu_id=338&action=479

houssine78 (approved):

beesdoo_product/models/res_conig_settings.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2019-2020 Elouan Le Bars <elouan@coopiteasy.be>
```
houssine78: file name misspelled

beesdoo_product/models/__init__.py
```diff
@@ -1 +1,2 @@
 from . import beesdoo_product
+from . import res_conig_settings
```
houssine78: file name misspelled

beesdoo_product/__manifest__.py
```diff
@@ -21,6 +21,7 @@
         "data/product_sequence.xml",
```
houssine78, line 22: change version

houssine78: add your name to the contributors

beesdoo_product/models/res_conig_settings.py
```diff
@@ -0,0 +1,41 @@
...
+from odoo import api, fields, models
+
+
+class ResConfigSettings(models.TransientModel):
+    _inherit = "res.config.settings"
+
+    suggested_price_reference = fields.Selection(
+        selection=[
+            ("supplier_price", "Supplier Price"),
+            ("sale_price", "Sale Price"),
```
houssine78: ("supplier_price", "On supplier Price"),
("sale_price", "On sale Price"),

beesdoo_product/models/res_conig_settings.py
```diff
@@ -0,0 +1,41 @@
...
+            - Margin on Supplier Price : Suggested sale price = supplier price * (1 + margin / 100) (default)
+            - Margin on Sale Price: Suggested sale price = supplier price * (1 / (1 - margin / 100))
+        """,
+        default="supplier_price",
+    )
+
+    @api.model
+    def get_values(self):
+        res = super(ResConfigSettings, self).get_values()
+        select_type = self.env["ir.config_parameter"].sudo()
```
houssine78: having sudo activated at this level is something that I dislike... I rather prefer having it on function call for security reason.

beesdoo_product/models/res_conig_settings.py
```diff
@@ -0,0 +1,41 @@
...
+        suggested_price_reference = select_type.get_param(
+            "beesdoo_product.suggested_price_reference"
+        )
+        res.update({"suggested_price_reference": suggested_price_reference})
+        return res
+
+    @api.multi
+    def set_values(self):
+        super(ResConfigSettings, self).set_values()
+        select_type = self.env["ir.config_parameter"].sudo()
```
houssine78: same here

beesdoo_product/models/beesdoo_product.py
```diff
@@ -247,16 +265,45 @@ def _unit_same_category(self):
...
+                    0
+                ].product_tmpl_id.categ_id.profit_margin
+                profit_margin = (
+                    profit_margin_supplier or profit_margin_product_category
+                )
+                suggested_price_reference = (
+                    self.env["ir.config_parameter"]
+                    .sudo()
+                    .get_param("beesdoo_product.suggested_price_reference")
+                )
```
houssine78: This should go befor the for loop as it does not depends on the product

beesdoo_product/models/beesdoo_product.py
```diff
@@ -247,16 +265,45 @@ def _unit_same_category(self):
...
+                    lambda t: t.amount_type == "percent" and t.price_include
+                )
+                sale_taxes_factor = 1 + sum(sale_taxes.mapped("amount")) / 100
+                profit_margin_supplier = suppliers[0].name.profit_margin
+                profit_margin_product_category = suppliers[
+                    0
+                ].product_tmpl_id.categ_id.profit_margin
+                profit_margin = (
+                    profit_margin_supplier or profit_margin_product_category
+                )
```
houssine78, line 295: Half of this code, including database call will never apply on a database considering suggested_price_reference. But we perform both options on each loop.

The code should go inside the if/else statement...

beesdoo_product/readme/DESCRIPTION.rst
```diff
@@ -1,2 +1,6 @@
 Modification of product module for the needs of beescoop
-- SOOO5 - Ajout de label bio/ethique/provenance
+- SOOO5 - Adds the label bio/ethique/provenance
+- Add a 'Suggested Price' field on products, and a 'Product Margin' field on Partners (Vendors) and Product Categories. The first margin is used if set, otherwise the second margin (which has a default value) is used.
```
houssine78, line 3: Just add a note about the assumption that we made for this module. We suppose that each product has only one supplier and that same product coming from different suppliers should be duplicated.

houssine78 (changes requested):
please take comments into accounts

beesdoo_product/models/res_conig_settings.py
mclaeysb: This is a standard block of code. It's like this, with the `sudo`, almost everywhere..

houssine78: It is not the main way of use. I know that in Odoo standard code we find it done that way sometimes mainly in res.config stuff but it's not recommended anywhere else.

I'm not against keeping it for res.config but a rather prefer standardization of use among our code. Maybe a point to discuss in coder council :)

beesdoo_product/models/beesdoo_product.py
mclaeysb: I think you're reasoning here is "depending on the value of `suggested_price_reference` either `supplier_taxes_factor` or `sale_taxes_factor` is computed and used."
It's actually: "depending on the value of `suggested_price_reference` the profit_margin_factor is computed differently from the profit_margin, and independent of that both `supplier_taxes_factor` or `sale_taxes_factor` are always computed and used."
So I don't think a change is required here.

houssine78: Ok I got it ! Could hence add some comment to make it easier to dig in ? thanks !

houssine78 (approved):

PR #165, opened 2020-08-02
[12.0][ADD] beesdoo_product: Add "FDS" and "Danger particulier" field

StCyr:
Implements request 155 in file "20181114_Planification tache Odoo coop it easy"
# Request
Ajouter 2 champs à la vue article pour gérer les composantes chimiques des produits et leur danger possible
- champ "FDS", avec valeurs possibles : non requise (valeur par défaut), requise, présente
- champ "Danger particulier", avec valeurs possibles : non (valeur par défaut), acide, base, autre danger
# Status
Working but still missing:
- Translation of the "FDS" and "Danger particulier" field's values

Some questions also:
- I've created a new model, BeesdooProductHazard, for this (copied from BeesdooProductLabel). Should I rather use BeesdooProductLabel?
- On my dev instance, inactive records are not shown in tree views by default. This has the result that when I mark a label inactive it disappear from the view. Is it working the same way in production? If so, can/should we change that (at least for the BeesdooProductLabel and BeesdooProductHazard model)?

### DIFF
--- a/beesdoo_product/models/beesdoo_product.py
+++ b/beesdoo_product/models/beesdoo_product.py
@@ -11,6 +11,19 @@ from odoo.tools.translate import _
 _logger = logging.getLogger(__name__)
 
 
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    profit_margin = fields.Float(string="Product Margin [%]")
+
+    @api.multi
+    @api.constrains("profit_margin")
+    def _check_margin(self):
+        for product in self:
+            if product.profit_margin < 0.0:
+                raise UserError(_("Percentages for Profit Margin must > 0."))
+
+
 class BeesdooProduct(models.Model):
     _inherit = "product.template"
 
@@ -252,12 +265,17 @@ class BeesdooProduct(models.Model):
         for product in self:
             suppliers = product._get_main_supplier_info()
             if len(suppliers) > 0:
-                product.suggested_price = (
-                    suppliers[0].price * product.uom_po_id.factor
-                ) * (
-                    1
-                    + suppliers[0].product_tmpl_id.categ_id.profit_margin / 100
+                price = suppliers[0].price
+                profit_margin_supplier = suppliers[0].name.profit_margin
+                profit_margin_product_category = suppliers[
+                    0
+                ].product_tmpl_id.categ_id.profit_margin
+                profit_margin = (
+                    profit_margin_supplier or profit_margin_product_category
                 )
+                product.suggested_price = (
+                    price * product.uom_po_id.factor
+                ) * (1 + profit_margin / 100)
 
 
 class BeesdooScaleCategory(models.Model):
--- a/beesdoo_product/models/beesdoo_product.py
+++ b/beesdoo_product/models/beesdoo_product.py
@@ -72,10 +72,8 @@ class BeesdooProduct(models.Model):
 
     note = fields.Text("Comments")
 
-    # S0023 : List_price = Price HTVA, so add a suggested price
-    list_price = fields.Float(string="exVAT Price")
     suggested_price = fields.Float(
-        string="Suggested exVAT Price", compute="_compute_cost", readOnly=True,
+        string="Suggested Price", compute="_compute_cost", readOnly=True,
         help="""
         This field computes a suggested price based on the 'Product Margin' 
         field on Partners (Vendors), if it's set, or otherwise on the 'Product 
@@ -332,7 +330,7 @@ class BeesdooProductCategory(models.Model):
 class BeesdooProductSupplierInfo(models.Model):
     _inherit = "product.supplierinfo"
 
-    price = fields.Float("exVAT Price")
+    price = fields.Float("Price")
 
 
 class BeesdooUOMCateg(models.Model):
--- a/beesdoo_product/models/beesdoo_product.py
+++ b/beesdoo_product/models/beesdoo_product.py
@@ -24,20 +24,53 @@ class ResPartner(models.Model):
                 raise UserError(_("Percentages for Profit Margin must >= 0."))
 
 
+class BeesdooProductHazard(models.Model):
+    _name = "beesdoo.product.hazard"
+    _description = "beesdoo.product.hazard"
+
+    name = fields.Char()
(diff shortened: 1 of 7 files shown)


beesdoo_product/models/beesdoo_product.py
```diff
@@ -27,6 +27,13 @@ class BeesdooProduct(models.Model):
         "beesdoo.product.label", domain=[("type", "=", "delivery")]
     )
 
+    fds_label = fields.Many2one(
+        "beesdoo.product.hazard", string="FDS label", domain=[("type", "=", "fds")], translate=True, default=lambda self: self.env['beesdoo.product.hazard'].search([["type", "=", "fds"],["name", "=", "Not required"]])
+    )
+    hazard_label = fields.Many2one(
+        "beesdoo.product.hazard", string="Hazard label", domain=[("type", "=", "hazard")], translate=True, default=lambda self: self.env['beesdoo.product.hazard'].search([["type", "=", "hazard"],["name", "=", "No"]])
+    )
+
```
houssine78: could you respect PEP8 convention and have one field attribute on each line

beesdoo_product/models/beesdoo_product.py
```diff
@@ -276,6 +283,20 @@ class BeesdooScaleCategory(models.Model):
...
+    _description = "beesdoo.product.hazard"
+
+    name = fields.Char()
+    type = fields.Selection(
+        [
+            ("fds", "FDS"),
+            ("hazard", "Specific hazard"),
+        ]
+    )
+    active = fields.Boolean(default=True)
```
houssine78: Why having active field ? I don't see the point for such case.

beesdoo_product/views/beesdoo_product.xml
```diff
@@ -110,6 +114,32 @@
...
+    </record>
+
+    <record model="ir.ui.view" id="beesdoo_product_hazard_tree">
+        <field name="name">bees.product.hazard.tree</field>
+        <field name="model">beesdoo.product.hazard</field>
+        <field name="arch" type="xml">
+            <tree editable="top">
+                <field name="name"/>
+                <field name="type"/>
+                <field name="active"/>
```
houssine78, line 138: if you want to keep no need to have it in a tree view as Odoo hide inactive records by the default

beesdoo_product/views/beesdoo_product.xml
```diff
@@ -45,6 +45,10 @@
                             <field name="label_last_printed"/>
                         </group>
                     </group>
+                    <group>
+                        <field name="fds_label"/>
+                        <field name="hazard_label"/>
```
houssine78: add widget="selection" or options="no_create" to avoid the user to create label on the fly from here.

houssine78:
I made some comments

for you question 

- "I've created a new model, BeesdooProductHazard, for this (copied from BeesdooProductLabel). Should I rather use BeesdooProductLabel?"
No better to have a dedicated model for this cause no way to distinguish all the different labels that can rapidly become a mess to manage.

- "On my dev instance, inactive records are not shown in tree views by default. This has the result that when I mark a label inactive it disappear from the view. Is it working the same way in production? If so, can/should we change that (at least for the BeesdooProductLabel and BeesdooProductHazard model)?"
this is standard in Odoo. It doesn't show the inactive record. First question, why would you have active field ? I don't see the point for changing the behaviour on production.

beesdoo_product/models/beesdoo_product.py
StCyr: Yes, absolutely: pretty code always works better :-)

beesdoo_product/models/beesdoo_product.py
StCyr: I just copy the functionality from the BeesdooProductLabel model. But, yeah, it doesn't work well

beesdoo_product/views/beesdoo_product.xml
StCyr: I don't want to keep the active field

beesdoo_product/views/beesdoo_product.xml
StCyr: Thanks for the tip (options="no_create" doesn't seem to work here though)

StCyr:
thanks for your reviexw @houssine78 

I've updated my PR related to your PEP8 and widget="selection" remarks

Regarding the active field, I'm happy with removing it. Though, I would like you to confirm this a second time as it is a functionality that's implemented in the BeesdooProductLabel model and that I've just copied from there. The question I'm asking is: "Why would an active field make sense in the BeesdooProductLabel model and not in the BeesdooProductHazard one?"

houssine78:
@StCyr just remove it. As people will disable the it and someone will create it after that. If there is a need to add it, we will do it afterward but I don't like to give this option at the beginning as most people make a wrong usage of it.

StCyr:
> @StCyr just remove it. As people will disable the it and someone will create it after that. If there is a need to add it, we will do it afterward but I don't like to give this option at the beginning as most people make a wrong usage of it.

En fait, ils veulent cette fonctionnalité: "10/07/2020 Virg/Bart : oui, comme eco label, origin label => les valeurs doivent être configurables/**archivables**."

Virginie l'a encore re-confirmé par email: "Je ne vois pas d'inconvénient à laisser la possibilité d'archiver les valeurs (le jour où les réglementations changent, de nouvelles valeurs seront peut-être requises et les anciennes à ne plus utiliser, donc archiver).
Pour ça peut rester ainsi."

houssine78:
@StCyr le client est roi ! enfin sauf en informatique :) bref laissons ça comme cela...

houssine78 (approved):

houssine78:
@StCyr for info the exact syntax is options="{'no_create': True}"

vvrossem:
Hi @StCyr ! Thank your for this PR :) 
I'm trying to deploy it on our testing server, but there are merge conflicts with another branch: `12.0-supplier_margin`
Could you kindly rebase your branch on `12.0-supplier_margin` ? 
Many thanks !

StCyr:
done

vvrossem:
Thank you, it's now under testing

PR #174, opened 2020-09-02
[12.0] fix main supplier selection to make it deterministic

robinkeunen:


purchase_order_generator/models/purchase_order_generator.py
```diff
@@ -84,8 +84,17 @@ def generate_cpo(self):
...
-                ]
+                ],
+                order="date_start desc",
+                limit=1,
             )
+
+            if not supplierinfo:
+                product_name = self.env["product.template"].browse(product_id)
+                raise ValidationError(
+                    _("No supplier defined for product %s") % product_name
```
houssine78, line 95: you should give the translated product name because they manly duplicate it and so only the translation is correct while the product name is the one from the duplicated name (with copy mentionned).

purchase_order_generator/models/purchase_order_generator.py
```diff
@@ -84,8 +84,17 @@ def generate_cpo(self):
...
                     ("product_tmpl_id", "=", product_id),
                     ("name", "=", supplier.id),
-                ]
+                ],
+                order="date_start desc",
+                limit=1,
             )
+
+            if not supplierinfo:
+                product_name = self.env["product.template"].browse(product_id)
```
houssine78: browsing in a for loop is bad for perf issue. you should loop on the browsed products

Doing something like for product in self.env["product.template"].browse(product_ids):

houssine78 (changes requested):
could you have a look at my comments ? thanks !

houssine78:
@robinkeunen any task linked to this PR ?

purchase_order_generator/models/purchase_order_generator.py
robinkeunen: But I browse only to display the product name in the error so I'm browsing at most once. :-)

purchase_order_generator/models/purchase_order_generator.py
robinkeunen: It's ok, we talked about it.

robinkeunen:
@houssine78 yes: https://gestion.coopiteasy.be/web#id=22067&view_type=form&model=account.analytic.line&menu_id=274&action=365

robinkeunen:
can you approve the review then? :-)

vvrossem (approved):
my 20cents : if `raise ValidationError` immediately breaks the for-loop, there shouldn't be a perf issue

houssine78 (approved):

houssine78:
> 
> 
> @houssine78 yes: https://gestion.coopiteasy.be/web#id=22067&view_type=form&model=account.analytic.line&menu_id=274&action=365

this is not the task

robinkeunen:
woops sorry:  https://gestion.coopiteasy.be/web#id=5083&view_type=form&model=project.task&menu_id=338&action=479

PR #179, opened 2020-09-09
[12.0][ADD] beesdoo_pos: available_in_pos management

vvrossem:
[Task](https://gestion.coopiteasy.be/web#id=5085&view_type=form&model=project.task&action=475&active_id=117)
This module maintains "Available in POS" (`available_in_pos`) value when unchecking "Can be Sold" (`sale_ok`).

### DIFF
--- a/beesdoo_pos/__manifest__.py
+++ b/beesdoo_pos/__manifest__.py
@@ -1,6 +1,7 @@
 # Copyright 2017 - 2020 BEES coop SCRLfs
 #   - Elouan Lebars <elouan@coopiteasy.be>
 #   - Rémy Taymans <remy@coopiteasy.be>
+#   - Vincent Van Rossem <vincent@coopiteasy.be>
 #   - Elise Dupont
 #   - Thibault François
 #   - Grégoire Leeuwerck
@@ -12,7 +13,7 @@
     "author": "Beescoop - Cellule IT, Coop IT Easy SCRLfs",
     "website": "https://github.com/beescoop/Obeesdoo",
     "category": "Point Of Sale",
-    "version": "12.0.1.0.0",
+    "version": "12.0.1.1.0",
     "depends": ["beesdoo_base", "beesdoo_product"],
     "data": ["views/beesdoo_pos.xml", "data/default_barcode_pattern.xml"],
     "qweb": ["static/src/xml/templates.xml"],
--- a/beesdoo_pos/models/product.py
+++ b/beesdoo_pos/models/product.py
@@ -1,0 +1,16 @@
+from odoo import api, fields, models, _
+
+
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    @api.onchange('sale_ok')
+    def _onchange_sale_ok(self):
+        """ Maintains the value of `available_in_pos` when unchecking `sale_ok`,
+            by storing its value before calling parent's method,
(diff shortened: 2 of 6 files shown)


beesdoo_pos/models/product.py
```diff
@@ -0,0 +1,16 @@
...
+class ProductTemplate(models.Model):
+    _inherit = 'product.template'
+
+    @api.onchange('sale_ok')
+    def _onchange_sale_ok(self):
+        """ Maintains the value of `available_in_pos` when unchecking `sale_ok`,
+            by storing its value before calling parent's method,
+            then assigning the stored value.
+        """
+        is_available = self.available_in_pos
```
houssine78, line 13: is_available is misleading I prefere to use was_available instead.

houssine78 (approved):

beesdoo_pos/__manifest__.py
```diff
@@ -12,7 +13,7 @@
     "author": "Beescoop - Cellule IT, Coop IT Easy SCRLfs",
     "website": "https://github.com/beescoop/Obeesdoo",
     "category": "Point Of Sale",
-    "version": "12.0.1.0.0",
+    "version": "12.0.1.0.1",
```
remytms: I suggest to increment the second digit of the version number as it adds new functionality in the module and not a fix in the current one. See https://semver.org
```suggestion
    "version": "12.0.1.1.0",
```

beesdoo_pos/models/product.py
```diff
@@ -0,0 +1,16 @@
...
+        super(ProductTemplate, self)._onchange_sale_ok()
+        if not self.sale_ok:
+            self.available_in_pos = is_available
```
remytms: Have you tested that a product with `sale_ok` False and `available_in_pos` True is realy not available in pos. My fear is that the POS may rely only in the `available_in_pos` value to check if the product can appear in the POS regardless of the value of `sale_ok`.

If you tested this case, this "naive" approach may be good. If not the previous value of `available_in_pos` should be stored in a specific field in database to be restored properly.

beesdoo_pos/readme/DESCRIPTION.rst
```diff
@@ -1 +1,3 @@
-This module adds the eaters of the customer to the POS ActionpadWidget and PaymentScreenWidget.
+This module :
+- adds the eaters of the customer to the POS ActionpadWidget and PaymentScreenWidget.
+- maintains "Available in POS" (`available_in_pos`) value when unchecking "Can be Sold" (`sale_ok`).
```
remytms, line 3: Thanks for updating readme.

remytms (changes requested):
Some things should be checked. :)

beesdoo_pos/models/product.py
vvrossem: I've tested the case : 

1. a product has `sale_ok = False` and `available_in_pos = True`
2. opened a POS session, said product doesn't appear 

=> a product appear in the POS seems to be depending on the value of `sale_ok`

beesdoo_pos/__manifest__.py
vvrossem: done :)

vvrossem:
All requested changed are now made

remytms (approved):
LGTM.

PR #181, opened 2020-09-09
[IMP] p_o_g: Filter product based on supplier

remytms:
See [task](https://gestion.coopiteasy.be/web#id=5136&view_type=form&model=project.task&action=479)


robinkeunen (approved):
Clean. Did not know about domains and onchange.

PR #189, opened 2020-09-28
[FIX] beesdoo_base: group_force_barcode (users eval)

vvrossem:
[Task](https://gestion.coopiteasy.be/web#id=5205&view_type=form&model=project.task)

- (6, _, [ids]) replaces the list of linked records with the provided list.
- (4, id, _) links an already existing record.This can only be used for many-to-many fields

### DIFF
--- a/beesdoo_base/security/groups.xml
+++ b/beesdoo_base/security/groups.xml
@@ -2,6 +2,6 @@
 <odoo>
     <record id="group_force_barcode" model="res.groups">
         <field name="name">Bees Card Force Barcode</field>
-        <field name="users" eval="[(6, 0, [ref('base.user_root')])]"/>
+        <field name="users" eval="[(4, ref('base.user_root'))]"/>
     </record>
 </odoo>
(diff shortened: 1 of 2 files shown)


beesdoo_base/security/groups.xml
```diff
@@ -2,6 +2,6 @@
 <odoo>
     <record id="group_force_barcode" model="res.groups">
         <field name="name">Bees Card Force Barcode</field>
-        <field name="users" eval="[(6, 0, [ref('base.user_root')])]"/>
+        <field name="users" eval="[(4, ref('base.user_root'))]"/>
```
robinkeunen, line 5: Shouldn't it be `[(4, ref('base.user_root'), False)]`  (triplet)?

https://www.odoo.com/documentation/12.0/reference/orm.html

robinkeunen (changes requested):

beesdoo_base/security/groups.xml
vvrossem: C'est optionnel:
> The `_` underscore symbol used in the preceding list represents irrelevant values, usually
filled with `0` or `False` .The trailing irrelevant values can be safely omitted. For example, `(4, id,_)` can be used as `(4, id)`.
source : odoo 12 development essentials, p.148-149

robinkeunen (approved):

beesdoo_base/security/groups.xml
robinkeunen: then 👌

PR #197, opened 2020-10-26
Translate Status field in beesdoo.shift.shift

cesarlr:


cesarlr:
Hey @mclaeysb, we need to translate one field.

robinkeunen (changes requested):
@cesarlr I do not see elsewhere in the code the use of `_` for selection fields. The string value is added to the translated termes by Odoo. Did you check

- in the translated terms?
- generating the missing terms?

cesarlr:
Thanks for the quick reply @robinkeunen! I have checked both but they are missing anyway. I think the problem is that the fields are not directly in selection but they are computed in a function. That's the reason I use underscore.

robinkeunen:
Yes you're right. Can you also adapt the code that way in beesdoo_shift/models.task.py? 

```python

    ##################################
    # Method to override             #
    # to have different state        #
    # on the shift                   #
    ##################################
    def _get_selection_status(self):
        return [
            ("open", "Confirmed"),
            ("done", "Attended"),
            ("absent", "Absent"),
            ("excused", "Excused"),
            ("cancel", "Cancelled"),
        ]
```

cesarlr:
sure!

robinkeunen (approved):

robinkeunen:
Thanks for the fix!

PR #198, opened 2020-11-03
[FIX] beesdoo_website_shift: shift_irregular_worker noupdate false

mclaeysb:
[Task](https://gestion.coopiteasy.be/web?token=R0Qs92cjQIn6BsLb5XTZ&db=coopiteasy#id=4925&view_type=form&model=project.task&action=479)

Set `noupdate = false` on the `shift_irregular_worker` pages.

### DIFF
--- a/beesdoo_website_shift/views/shift_website_templates.xml
+++ b/beesdoo_website_shift/views/shift_website_templates.xml
@@ -6,7 +6,7 @@
 <odoo>
 
     <!-- Add menu entries -->
-    <data noupdate="1">
+    <data noupdate="0">
         <record id="menu_work_irregular" model="website.menu">
             <field name="name">Shifts Irregular</field>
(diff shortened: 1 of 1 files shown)


robinkeunen (approved):

beesdoo_website_shift/views/shift_website_templates.xml
```diff
@@ -6,7 +6,7 @@
 <odoo>
 
     <!-- Add menu entries -->
-    <data noupdate="1">
+    <data noupdate="0">
```
houssine78, line 9: when it's 0 you can remove noupdate parameter. I think that you can even remove data tags as it's no more needed on 12.0 if not wanting to have noupdate set to True

remytms:
We should keep, `noupdate=1` on menu record. It should not be updated if a customer remove it from it's website. We can always create it again via the backend interface. :)

PR #201, opened 2020-11-17
[12.0][IMP] beesdoo_purchase: button to adapt purchase/selling price

vvrossem:
[Task](https://gestion.coopiteasy.be/web#id=5545&view_type=form&model=project.task)
**Before**
Purchase and/or selling price was adapted when confirming PO.

**After**
Purchase and/or selling price is now adapted when using button "Adapt Prices".
- button is visible only for the following states: `draft,sent,purchase`
- replaced toggle buttons with select/deselect all
- performance: loops only the lines with `adapt_[purchase|selling]_price` set
- raise an UserError if no seller (`product.supplierinfo`) is found
  when using standard method `_select_seller` from `product.product`.
  A reason could be that the product's ordered `quantity` is lower than
  the seller `min_qty`

### DIFF
--- a/beesdoo_purchase/__manifest__.py
+++ b/beesdoo_purchase/__manifest__.py
@@ -1,15 +1,13 @@
 {
     "name": "Bees Purchase",
     "summary": """
-        - Adds a 'Responsible' field to purchase orders
-        - A filter w.r.t. the mail sellers is placed on the products field of a
-        purchase order
-        - Allow inverting the Purchase Order Reference on the invoice lines
+        Enhancements related to Purchase module :
+        field, filter, PO reference, product's purchase and/or selling price
     """,
     "author": "Beescoop - Cellule IT, " "Coop IT Easy SCRLfs",
     "website": "https://github.com/beescoop/Obeesdoo",
     "category": "Purchase",
-    "version": "12.0.1.2.0",
+    "version": "12.0.1.3.0",
     "depends": ["base", "purchase", "beesdoo_product"],
     "data": [
         "security/invoice_security.xml",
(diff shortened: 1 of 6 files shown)


beesdoo_purchase/__manifest__.py
```diff
@@ -5,11 +5,12 @@
         - A filter w.r.t. the mail sellers is placed on the products field of a
         purchase order
         - Allow inverting the Purchase Order Reference on the invoice lines
+        - Allows adapting a product's purchase and/or selling price from a PO.
```
robinkeunen: This makes the summary entry too long and breaks the module loading on any server that has this module  available. No less :p

So this should be summarized into a one liner since it's redundant w/ DESCRIPTION.rst

beesdoo_purchase/models/purchase.py
```diff
@@ -44,42 +45,49 @@ def write(self, vals):
...
+                        # on `product.template` in `beesdoo_product`
+                        seller.price = price
+                        product_tmpl_id.list_price = suggested_price
+                    elif line.adapt_purchase_price:
+                        seller.price = price  # see above comment
+                    elif line.adapt_selling_price:
+                        product_tmpl_id.list_price = suggested_price
+                else:
+                    raise UserError(
+                        _("Cannot adapt the price of '%s'.") % product_id.name
```
robinkeunen: Maybe more explicit?

No seller found, Odoo cannot adapt the price of '%s'

robinkeunen (changes requested):

beesdoo_purchase/__manifest__.py
vvrossem: done :)

beesdoo_purchase/models/purchase.py
vvrossem: done :)

robinkeunen (approved):

PR #202, opened 2020-12-03
[ADD] beesdoo_product_pricelist

vvrossem:
## [Task](https://gestion.coopiteasy.be/web#id=5277&view_type=form&model=project.task)

Adds "Edit Price" submenu on Point Of Sale, Purchase and Sale modules.
The user lands on an editable List View with the following columns :

- Name (`name`)
- Main Seller (`main_seller_id`)
- Purchase Price (`purchase_price`)
- Purchase Unit of Measure (`uom_po_id`)
- Suggested Price (`suggested_price`)
- Sales Price (`list_price`)
- Unit of Measure (`uom_id`)

The only editable field is Purchase Price.

Through "Action > Adapt Sales Price", the user can, on the selected products, adapt the Sales Price according to the Suggested Price.


beesdoo_product_pricelist/__manifest__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright 2020 - BEES coop SCRLfs
```
houssine78, line 1: je mettrai coopiteasy pour le copyright vu que cela ne se fait pas en bénévolat dans la cellule de la bees

beesdoo_product_pricelist/models/product_template.py
```diff
@@ -0,0 +1,33 @@
...
+    @api.multi
+    def _inverse_purchase_price(self):
+        for product in self:
+            supplierinfo = product._get_main_supplier_info()
+            if supplierinfo:
+                supplierinfo.price = product.purchase_price
+            else:
+                raise ValidationError(
+                    _("No Vendor defined for product '%s'") % product.name
+                )
```
houssine78, line 33: I think that this code should go in beesdoo_product module

houssine78 (approved):
I think that all the code should go in beesdoo_product as it doesn't bring a different behavior but rather bring additional menu and functionalities which are not forced to be used.

It would be easier to maintain on module than two.

robinkeunen:
@houssine78 I'm not sure it should be moved. We've been asked by coopdev to split the modules so that it's easier for them to adapt to their context (GRAP also made us similar requests for odoo-cae). Plus, I think it's easier to maintain and migrate small functional modules rather than a big fat module.

beesdoo_product_pricelist/views/product_template_views.xml
```diff
@@ -0,0 +1,22 @@
...
+        <field name="name">product.template.edit.price.tree</field>
+        <field name="model">product.template</field>
+        <field name="arch" type="xml">
+            <tree string="Product" editable="top" edit="true">
+                <field name="name" readonly="1"/>
+                <field name="main_seller_id" string="Main Seller" readonly="1"
+                       options="{'no_open': True, 'no_create': True, 'no_create_edit': True }"/>
+                <field name="purchase_price"/>
+                <field name="uom_po_id" readonly="1"
+                       options="{'no_open': True, 'no_create': True, 'no_create_edit':True }"/>
```
robinkeunen, line 14: 👍

houssine78:
@robinkeunen I would agree with you if we were talking about module like shift or emc but beesdoo product is pretty light and simple. And this module has a dependency on several beesdoo modules (not sure they're all needed btw) that's why I'm suggesting to integrate it in beesdoo_product rather than have one more module for so small and no generic stuff

beesdoo_product_pricelist/__manifest__.py
```diff
@@ -0,0 +1,23 @@
...
+{
+    "name": "beesdoo_product_pricelist",
+    "summary": """
+        Allow quick product price editing
+        on point of sale, purchase and sale modules""",
+    "author": "Coop IT Easy SCRLfs",
+    "website": "https://github.com/beescoop/Obeesdoo",
+    "category": "Sales",
+    "version": "12.0.1.0.0",
+    "depends": ["beesdoo_product", "point_of_sale", "purchase", "sale"],
```
houssine78, line 13: @vvrossem do the dependency on beesdoo product is necessary. if we go with the suggestion of @robinkeunen I would rather see having beesdoo_product depending on this one (As far as it's possible).

vvrossem: It's necessary because `beesdoo_product` declares the field `main_seller_id`

robinkeunen (approved):

robinkeunen:
I do agree that the dependency on sale + pos + purchase makes a good point to include this dev into beesdoo_product.

robinkeunen:
What does Coopdev think about it? @cesarlr @enricostano

vvrossem:
closing in favor of https://github.com/beescoop/Obeesdoo/pull/203 that includes this "quick price editing" feature in `beesdoo_product`

PR #212, opened 2021-01-19
[FIX]: add case to compute stock coverage when stock is empty

leeuwerck:
https://gestion.coopiteasy.be/web#id=5590&view_type=form&model=project.task&action=479

### DIFF
--- a/beesdoo_purchase/__manifest__.py
+++ b/beesdoo_purchase/__manifest__.py
@@ -7,8 +7,13 @@
     "author": "Beescoop - Cellule IT, " "Coop IT Easy SCRLfs",
     "website": "https://github.com/beescoop/Obeesdoo",
     "category": "Purchase",
-    "version": "12.0.1.3.0",
-    "depends": ["base", "purchase", "beesdoo_product"],
+    "version": "12.0.1.4.0",
+    "depends": [
+        "base",
+        "purchase",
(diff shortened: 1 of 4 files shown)


beesdoo_purchase/__manifest__.py
```diff
@@ -8,7 +8,12 @@
     "website": "https://github.com/beescoop/Obeesdoo",
     "category": "Purchase",
     "version": "12.0.1.3.0",
```
robinkeunen: increment minor version

beesdoo_purchase/models/purchase.py
```diff
@@ -120,3 +120,15 @@ class PurchaseOrderLine(models.Model):
         help="Check this box to adapt the selling price "
         "on the product page when confirming Purchase Order",
     )
+
+    stock_coverage = fields.Float(compute="_compute_product", store=True)
+
+    @api.depends("product_id")
+    def _compute_product(self):
```
robinkeunen: `_compute_stock_coverage` would probably be a better name.

robinkeunen (changes requested):
A few comments I could have told directly I guess :p

I'll fix travis, there might be a few tests to fix.

robinkeunen (approved):

PR #216, opened 2021-02-10
[ADD] beesdoo purchase: add option for rounding suggested price

leeuwerck:
https://gestion.coopiteasy.be/web#id=5677&view_type=form&model=project.task&action=479

### DIFF
--- a/beesdoo_product/models/beesdoo_product.py
+++ b/beesdoo_product/models/beesdoo_product.py
@@ -7,6 +7,7 @@ from datetime import date
 
 from odoo import api, fields, models
 from odoo.exceptions import UserError, ValidationError
+from odoo.tools import float_round
 from odoo.tools.translate import _
 
 _logger = logging.getLogger(__name__)
@@ -37,9 +38,7 @@ class BeesdooProductHazard(models.Model):
 class BeesdooProduct(models.Model):
     _inherit = "product.template"
 
-    eco_label = fields.Many2one(
-        "beesdoo.product.label", domain=[("type", "=", "eco")]
-    )
+    eco_label = fields.Many2one("beesdoo.product.label", domain=[("type", "=", "eco")])
     local_label = fields.Many2one(
         "beesdoo.product.label", domain=[("type", "=", "local")]
     )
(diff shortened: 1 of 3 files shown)


beesdoo_product/models/beesdoo_product.py
```diff
@@ -340,6 +312,9 @@ def _compute_cost(self):
                     * profit_margin_factor
                 )
 
+                if suppliers[0].product_tmpl_id.categ_id.should_round_suggested_price:
+                    product.suggested_price = product.suggested_price
```
robinkeunen: I don't see any rounding here 🤔

beesdoo_product/views/beesdoo_product.xml
```diff
@@ -167,6 +167,7 @@
         <field name="arch" type="xml">
             <field name="display_name" position="after">
                 <field name="profit_margin"></field>
+                <field name="should_round_suggested_price"></field>
```
robinkeunen: Fais pas le souillon comme le type qui a mit la balise `profit_margin`:

```
<field name="should_round_suggested_price"></field>
should be
<field name="should_round_suggested_price"/>
```

beesdoo_product/views/beesdoo_product.xml
```diff
@@ -178,6 +179,7 @@
         <field name="arch" type="xml">
             <field name="parent_id" position="after">
                 <field name="profit_margin"></field>
+                <field name="should_round_suggested_price"></field>
```
robinkeunen: same here

beesdoo_product/models/beesdoo_product.py
```diff
@@ -430,3 +408,7 @@ class BeesdooUOMCateg(models.Model):
         string="Category type",
         default="unit",
     )
+
+
+def round_5c(price):
+    return float_round(price, precision_rounding=0.05, rounding_method="HALF")
```
robinkeunen, line 414: Put utility functions at the top of the file.

robinkeunen (changes requested):
See comments in the code.

Did you round with the same method as the sale price?

best practice: when you blacken code that is not part of the PR, try doing it in a separate commit so that the reviewer can focus on meaningful code. 
(does not happen with a working pre-commit)

robinkeunen (approved):

PR #218, opened 2021-03-04
[IMP] beesdoo_shift: unique status term for cooperative status

robinkeunen:
[T3381 - Impossible de trouver le terme source pour "Avertissement" (statut Alerte)](https://gestion.coopiteasy.be/web#id=6083&view_type=form&model=project.task)

Since the terms of these status are translated as "code" type          if they are already translated elsewhere, it won't be possible         to translate them to a specific status.

Hence these unique terms that have to be translated at module         configuration.

### DIFF
--- a/beesdoo_shift/README.rst
+++ b/beesdoo_shift/README.rst
@@ -26,6 +26,20 @@ Generate and manage shifts for cooperators.
 .. contents::
    :local:
 
+Configuration
+=============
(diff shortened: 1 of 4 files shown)


remytms (approved):
Peut-être mettre à jour les fichiers PO du coup ? Sinon, c'est cool !

PR #219, opened 2021-03-05
[IMP] make due date editable when a payment term is set

houssine78:
task explained [here](https://gestion.coopiteasy.be/web#id=6233&view_type=form&model=project.task&menu_id=338&action=479)

### DIFF
--- a/beesdoo_account/views/account_invoice.xml
+++ b/beesdoo_account/views/account_invoice.xml
@@ -18,9 +18,8 @@
             <field name="date_invoice" position="attributes">
                 <attribute name="required">True</attribute>
             </field>
-            <!-- doing it by using position="attributes" don't want to do the job.
-                 So I used with my muscles with replace.
-            -->
+            <!-- editable due date, event if payment_term is set;
+                 doing it by using position="attributes" does not do the job.-->
             <field name="date_due" position="replace">
                 <field name="date_due" attrs="{'readonly': [('state', 'in', ['open', 'in_payment', 'paid'])]}" force_save="1"/>
             </field>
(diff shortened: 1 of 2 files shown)


beesdoo_account/views/account_invoice.xml
```diff
@@ -18,6 +18,12 @@
             <field name="date_invoice" position="attributes">
                 <attribute name="required">True</attribute>
             </field>
+            <!-- doing it by using position="attributes" don't want to do the job.
+                 So I used with my muscles with replace.
```
vvrossem: what muscles ? :trollface:

beesdoo_account/views/account_invoice.xml
```diff
@@ -18,6 +18,12 @@
...
+            -->
+            <field name="date_due" position="replace">
+                <field name="date_due" attrs="{'readonly': [('state', 'in', ['open', 'in_payment', 'paid'])]}" force_save="1"/>
```
vvrossem, line 24: TIL : `force_save` send data to server for `readonly` fields

vvrossem (changes requested):
LGTM but I would remove your funny muscles joke :D

robinkeunen (commented):
Shouldn't we move this in bees_custom? Would all supermarkets  want this?

houssine78:
@robinkeunen I think that this will be useful for all the supermarket. But I'm ok with the proposition to put it in a stand alone module as it may interest other customers that have nothing to do with cooperative supermarket. It could also go to OCA

robinkeunen:
@houssine78 my last commit prevents odoo from resetting date_due to a computed value upon invoice validation (cf feedback from Virginie). 

However, I think it's not optimal since it prevents computation of date_due on a second change of payment_term_id. I don't think it's necessary to dig deeper in order to find a better way but that pushes the option to keep this out of beesdoo_account.

houssine78:
@robinkeunen ok for me ! I can't approve my own PR so consider this comment as an approval.

robinkeunen:
closed in favor of https://github.com/coopiteasy/cie-custom/pull/17

PR #225, opened 2021-03-20
Add a label_is_outdated field to product.template object.

Fkawala:
No tests made on this code b/c dockerfile is broken.

### DIFF
--- a/beesdoo_product/models/beesdoo_product.py
+++ b/beesdoo_product/models/beesdoo_product.py
@@ -92,6 +92,9 @@ class BeesdooProduct(models.Model):
 
     label_to_be_printed = fields.Boolean("Print label?")
     label_last_printed = fields.Datetime("Label last printed on")
+    label_is_outdated = fields.Boolean("label_is_outdated",
+        compute="_compute_label_is_outdated")
+
 
     note = fields.Text("Comments")
 
@@ -360,6 +363,10 @@ class BeesdooProduct(models.Model):
                 raise ValidationError(
                     _("No Vendor defined for product '%s'") % product.name
                 )
+
+    @api.multi
+    def _compute_label_is_outdated(self):
+        return self.write_date > self.label_last_printed
 
 
 class BeesdooScaleCategory(models.Model):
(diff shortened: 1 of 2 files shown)


robinkeunen (commented):
Hi @Fkawala ,

- what will you use this field for?
- I don't see any views and this is not stored, do we really need a field for this?
- isn't this condition too broad? Any change to any field will flag the product as outdated, you might want to look at some more specific fields?


(we should probably remove this docker file started by a bees volonteer, it never worked)

Fkawala:
Thank you for the comments Robin ! 
- The field is meant to be used to filter out products for which user needs to reprint labels.
- I'll add the search view I plan to have using this field.
- I agree this could be too broad (but as virtually any field (apart from comments?) of the object could be printed on the label I see no better option)

robinkeunen:
If you need to use the field in a search view, you'll need to store it or write a search function for the field :-)

In the future, we might analyse and refactor the way we print stickers. At that time we might move these fields around a bit (to custom modules?). It the meantime, let's go ahead with `label_is_outdated` once you contribute a search view.

Fkawala:
Thank you for the heads-up, will update to store the Be field. 

I'm not competent enough to comment your plan to move things around :) but I'm confident that's a smart strategy!

Fkawala:
@robinkeunen I added a filter based on the new field. Not sure this is what you had in mind though. Let me know & have a good day!

robinkeunen:
@Fkawala That's not exactly it. A computed field is not searchable if it is not either stored or is provided a search **function**. cf [odoo documentation here](https://www.odoo.com/documentation/12.0/reference/orm.html#computed-fields).

As you can see in the print screen here, the search **view** you provided does not appear because of that. More over, the field is not available in the _Add custom filter_ dropdown list. 

Screenshot 2021-03-29 at 11 33 41
Screenshot 2021-03-29 at 11 34 15
Screenshot 2021-03-29 at 11 38 43

Fkawala:
@robinkeunen Thank you for the detailed explanation. The commit https://github.com/beescoop/Obeesdoo/pull/225/commits/69aa1b13e715c55f93d1edeefc313a11f825dd49 should fix the problem you mentionned. Sorry if there is still an unsolved issue.

PR #244, opened 2021-06-21
[WIP] [CHG] b_shift: planning generation more generic

remytms:
This change does not change the algorithm of planning generation. But it changes what functions returns to let them be used elsewhere and be overwritten in other modules.

Credits to @elioexe


MatteoLdx:
https://github.com/beescoop/Obeesdoo/pull/311

PR #251, opened 2021-07-14
[12.0][ADD] beesdoo_shift: parameters to _cron_send_weekly_emails

vvrossem:
## [Task](https://gestion.coopiteasy.be/web#id=6478&view_type=form&model=project.task&action=479)

This PR adds the following parameters to both record `ir_cron_send_weekly_emails` and its method `_cron_send_weekly_emails`:
- `notice=1`
- `period=7`

As the record `ir_cron_send_weekly_emails` is in a `<data noupdate="1">` element, administrators need to first delete the record, then update the module in order to recreate it (if `beesdoo_shift` is already installed on a database)

However, with the default values, this modification is backward compatible and will not raise any error if the delete-the-record-then-update-the-module action is not done.

Special thanks to @Yo-Be for coming up with that solution

### DIFF
--- a/beesdoo_shift/data/cron.xml
+++ b/beesdoo_shift/data/cron.xml
@@ -38,7 +38,7 @@
             <field name="name">Send weekly shift summary</field>
             <field name="model_id" ref="model_beesdoo_shift_shift" />
             <field name="state">code</field>
-            <field name="code">model._cron_send_weekly_emails()</field>
+            <field name="code">model._cron_send_weekly_emails(notice=1, period=7)</field>
             <field name="interval_number">7</field>
             <field name="interval_type">days</field>
             <field name="numbercall">-1</field>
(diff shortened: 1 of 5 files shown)


beesdoo_shift/data/cron.xml
```diff
@@ -38,7 +38,7 @@
             <field name="name">Send weekly shift summary</field>
             <field name="model_id" ref="model_beesdoo_shift_shift" />
             <field name="state">code</field>
-            <field name="code">model._cron_send_weekly_emails()</field>
+            <field name="code">model._cron_send_weekly_emails(notice=1, period=7)</field>
             <field name="interval_number">7</field>
```
robinkeunen, line 42: I would add in the readme of the module that the interval number of the cron should be consistent with notice and period.

robinkeunen (changes requested):
As pointed, I think the PR should add a note in the readme.

I understand that another task is scheduled to make this configurable, if so, good for me when readme is added.

vvrossem:
README.rst updated with : 
```
- Configure "Send weekly shift summary" cron:

  - Its interval number (default = 7) should be consistent with notice (default = 1) and period (default = 7).
```

robinkeunen (approved):

PR #256, opened 2021-07-27
[12.0][FIX] suggested price computations

robinkeunen:
## ~~[T4322 - SPP - Erreur d'arrondi sur le prix de vente](https://gestion.coopiteasy.be/web#id=7024&view_type=form&model=project.task)~~

**Before**: `profit_margin_factor` not rounded leads to suggested price being too high
**After**: round `profit_margin_factor` to two digits~~

## [T4325 - SPP - Autre souci avec le calcul du prix de vente - unités](https://gestion.coopiteasy.be/web#id=7027&view_type=form&model=project.task)

**Before**: only `uom_po_id.factor` is considered when computing suggested price
**After**: compute ratio between `uom_po_id.factor` and `uom_id.factor`.

## Refactoring
- Blacken file
- `product._get_main_supplier_info()` always returns singleton or empty record set => do not index suppliers.

### DIFF
--- a/beesdoo_product/models/beesdoo_product.py
+++ b/beesdoo_product/models/beesdoo_product.py
@@ -38,7 +38,9 @@ class BeesdooProductHazard(models.Model):
 class BeesdooProduct(models.Model):
     _inherit = "product.template"
 
-    eco_label = fields.Many2one("beesdoo.product.label", domain=[("type", "=", "eco")])
+    eco_label = fields.Many2one(
+        "beesdoo.product.label", domain=[("type", "=", "eco")]
+    )
     local_label = fields.Many2one(
         "beesdoo.product.label", domain=[("type", "=", "local")]
     )
@@ -71,7 +73,9 @@ class BeesdooProduct(models.Model):
 
     display_unit = fields.Many2one("uom.uom")
     default_reference_unit = fields.Many2one("uom.uom")
-    display_weight = fields.Float(compute="_compute_display_weight", store=True)
+    display_weight = fields.Float(
+        compute="_compute_display_weight", store=True
+    )
 
     total_with_vat = fields.Float(
         compute="_compute_total",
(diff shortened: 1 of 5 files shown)


vvrossem (approved):

beesdoo_product/models/beesdoo_product.py
```diff
@@ -272,18 +299,28 @@ def _unit_same_category(self):
...
         suggested_price_reference = (
             self.env["ir.config_parameter"]
             .sudo()
             .get_param("beesdoo_product.suggested_price_reference")
         )
         for product in self:
-            suppliers = product._get_main_supplier_info()
-            if len(suppliers) > 0:
-                price = suppliers[0].price
+            supplier = product._get_main_supplier_info()
```
huguesdk, line 326: could `product.main_seller_id` be used instead? it seems that “seller” and “supplier” are used interchangeably in this class. is this on purpose?

looking at `_get_main_supplier_info()`, it took a little time to understand the logic. a “seller” with no `.date_start` comes before the others? tip in that function: instead of defining `far_future` with a magic date, use `date.max` directly. also, this function either returns a supplier or a collection. wouldn’t it be more logical to return a supplier or `None`?

beesdoo_product/models/beesdoo_product.py
```diff
@@ -272,18 +299,28 @@ def _unit_same_category(self):
...
+            if supplier:
```
huguesdk, line 327: i would clearly favor an early exit here. 😉
```python
if not supplier:
    continue
```
of course, this would make the whole block appear in the diff.

beesdoo_product/tests/test_suggested_price.py
```diff
@@ -0,0 +1,99 @@
...
+        self.env["product.supplierinfo"].create(
+            {
+                "name": self.supplier.id,
+                "price": 1.55,
+                "product_tmpl_id": product.product_tmpl_id.id,
+            }
+        )
+        self.assertEqual(
+            float_compare(product.suggested_price, 5.3, precision_digits=3), 0,
+        )
```
huguesdk, line 99: tests, yay! 🎉

beesdoo_product/models/beesdoo_product.py
robinkeunen: One point for you!

beesdoo_product/models/beesdoo_product.py
robinkeunen: > could product.main_seller_id be used instead? it seems that “seller” and “supplier” are used interchangeably in this class. is this on purpose?

Indeed, it would need a little cleanup. I left a comment to deal with it in the upcoming refactor done for PYSI.

> looking at _get_main_supplier_info(), it took a little time to understand the logic. a “seller” with no .date_start comes before the others?

Yes, I guess that comes from a specification that would have gained to be in a code comment. I added a comment.

> tip in that function: instead of defining far_future with a magic date, use date.max directly.

TIL 👍 I fixed it

> also, this function either returns a supplier or a collection. wouldn’t it be more logical to return a supplier or None?

Not sure about that. Another way to see it : the functions always returns a RecordSet : a RecordSet of 1 `supplierinfo` or an empty RecordSet. I left a comment in the code anyway.

I won't fix further since this is an old task that has been tested by Virginie and Pol. But I left comment that could be delt w/ in PYSI.

PR #258, opened 2021-08-05
[12.0][FIX] beesdoo_shift: working mode is not properly defined

vvrossem:
## [Task](https://gestion.coopiteasy.be/web#id=7081&view_type=form&model=project.task)

### DIFF
--- a/beesdoo_shift/wizard/subscribe.py
+++ b/beesdoo_shift/wizard/subscribe.py
@@ -199,10 +199,6 @@ class Subscribe(models.TransientModel):
             # ,...
             status_id.sudo().write(data)
 
-        if self.working_mode:
-            self.cooperator_id.sudo().write({
-                "working_mode": self.working_mode
-            })
         if self.shift_id:
             # Remove existing shift then subscribe to the new shift
             self.cooperator_id.sudo().write(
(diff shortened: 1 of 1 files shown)


beesdoo_shift/wizard/subscribe.py
```diff
@@ -207,4 +198,13 @@ def subscribe(self):
...
             status_id.sudo().write(data)
+
+        if self.shift_id:
+            # Remove existing shift then subscribe to the new shift
+            self.cooperator_id.sudo().write(
+                {"subscribed_shift_ids": [(6, 0, [self.shift_id.id])]}
+            )
+        if self.working_mode != "regular":
+            # Remove existing shift then subscribe to the new shift
+            self.cooperator_id.sudo().write({"subscribed_shift_ids": [(5,)]})
```
robinkeunen, line 209: @vvrossem Is this to cover the case when an irregular worker becomes a regular worker ? If so, this test should remain before the write / create.

robinkeunen (changes requested):

beesdoo_shift/wizard/subscribe.py
vvrossem: no, it's to fix a newly bug : error message "working mode is not correctly defined" when trying to subscribe a member to a shift and defining first the working mode

steps to reproduce the bug were :

- go on a cooperator form (Contacts)

- no working mode, no shift assigned

- click on button Subscribe to shift

- choose : Working mode = Regular worker, Choose Shift = X_Day_Time_Item

- Confirm => Error is raised: Working mode is not properly defined. Please check if the worker is subscribed

PR #261, opened 2021-12-02
[ADD] beesdoo_product: search product_code of main_seller_id

remytms:
[Task](https://gestion.coopiteasy.be/web#id=7693&view_type=form&model=project.task&action=475&active_id=213)

The purpose is to allow easy searching product when you have only the Vendor reference of the product (also known as `product_code` on the `product.supplierinfo` object).


beesdoo_product/README.rst
```diff
@@ -56,10 +60,10 @@ Please note that this model makes assumptions when computing the suggested price
 Bug Tracker
 ===========
 
-Bugs are tracked on `GitHub Issues <https://github.com/beescoop/obeesdoo/issues>`_.
+Bugs are tracked on `GitHub Issues <https://github.com/beescoop/addons/issues>`_.
 In case of trouble, please check there if your issue has already been reported.
 If you spotted it first, help us smashing it by providing a detailed and welcomed
-`feedback <https://github.com/beescoop/obeesdoo/issues/new?body=module:%20beesdoo_product%0Aversion:%2012.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
+`feedback <https://github.com/beescoop/addons/issues/new?body=module:%20beesdoo_product%0Aversion:%2012.0%0A%0A**Steps%20to%20reproduce**%0A-%20...%0A%0A**Current%20behavior**%0A%0A**Expected%20behavior**>`_.
```
robinkeunen: I think you generated the readme with the wrong `--repository` argument.

beesdoo_product/models/beesdoo_product.py
```diff
@@ -70,6 +70,11 @@ class BeesdooProduct(models.Model):
         compute="_compute_main_seller_id",
         store=True,
     )
+    main_seller_id_product_code = fields.Char(
+        string="Main Seller Product Code",
+        compute="_compute_main_seller_id",
+        store=True,
+    )
```
robinkeunen: That's a good way to do it. Maybe a better way would be to compute the top supplierinfo and have `main_seller_id` and `main_seller_id_product_code ` be related fields ?

```py
top_supplieinfo_id = fields.Many2one(
	comodel_name="product.supplierinfo", 
	compute=`_compute_main_seller_id`",
	store=True,
)
main_seller_id = fields.Many2one(
	comodel_name="res.partner", 
	related="top_supplieinfo_id.name
)
seller_product_code = fields.Char(
	related="top_supplierinfo_id.product_code"
)
```

What do you think about it ? Maybe add a todo ? Refactor it in Komunigi?

beesdoo_product/tests/test_search_main_seller_product_code.py
```diff
@@ -0,0 +1,56 @@
+# Copyright 2021 Coop IT Easy SCRL fs
+#   Rémy Taymans <remy@coopiteasy.be>
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+
+from odoo.tests import TransactionCase
+
+
+class TestSearchMainSellerProductCode(TransactionCase):
```
robinkeunen, line 9: 🙏

beesdoo_product/tests/test_search_main_seller_product_code.py
```diff
@@ -0,0 +1,56 @@
...
+
+        self.product1_main_supplierinfo = get_main_supplier_info(self.product1)
+        self.product2_main_supplierinfo = get_main_supplier_info(self.product2)
+
+    def test_search_product_by_main_seller_product_code(self):
+        """
+        Test searching a product giving the product_code of it's
+        main_seller_id
+        """
+        products = self.env["product.product"].search(
```
robinkeunen, line 25: I think this search does not match the GUI search. Here you can search on computed non-stored fields but not in the GUI. But I may be wrong. And I like tests.

robinkeunen: If your goal was rather to test your compute function, I'd name the test function differently.

robinkeunen (changes requested):

beesdoo_product/models/beesdoo_product.py
remytms: Would be nice indeed. I will try to do it now. :)

beesdoo_product/tests/test_search_main_seller_product_code.py
remytms: Ok, will change the name. Not sure how to test the GUI search. I was thinking that is use simply the `search` function.

remytms: In fact, my function was really to test that the `search` function works well on the related (previously computed) field. May be another fonction can be written to test the compute function.

beesdoo_product/README.rst
remytms: Indeed. Fixed.

robinkeunen (approved):
cleaner code, one fix at a time 👏

robinkeunen:
@remytms I think it's worth checking again that the filters do work on related fields, you might [have to store them](https://www.odoo.com/documentation/12.0/developer/reference/orm.html#related-fields).

robinkeunen (dismissed):
@remytms I think it's worth checking again that the filters do work on related fields, you might [have to store them](https://www.odoo.com/documentation/12.0/developer/reference/orm.html#related-fields).

robinkeunen:
@remytms This branch conflicts with the base branch. I have to remove it from gitaggregator.

carmenbianca:
temporarily removed this from repos.yml file because of conflict with #320

robinkeunen:
@remytms working on holiday ? Can we close or fix this ?

victor-champonnois:
~~removing from cie-repositories because of conflicts~~
nevermind, the issue was on my side

victor-champonnois:
@mieg Do you know if this can be merged ? It is blocking me with beesdoo_product refactoring. :)

remytms:
@victor-champonnois Can you review and merge. So that it can be deployed Monday. :)

remytms:
Just check to squash the fixup commit.

beesdoo_product/models/beesdoo_product.py
```diff
@@ -64,12 +64,20 @@ class BeesdooProduct(models.Model):
         translate=True,
     )
 
-    main_seller_id = fields.Many2one(
-        "res.partner",
-        string="Main Seller",
+    top_supplierinfo_id = fields.Many2one(
+        comodel_name="product.supplierinfo",
```
victor-champonnois, line 68: Why not `main_supplierinfo_id` ? It would be more consistent no ? Is there a reason for using "top" ?

beesdoo_product/models/beesdoo_product.py
```diff
@@ -64,12 +64,20 @@ class BeesdooProduct(models.Model):
...
         compute="_compute_main_seller_id",
         store=True,
     )
+    main_seller_id = fields.Many2one(
+        string="Main Seller",
+        comodel_name="res.partner",
+        related="top_supplierinfo_id.name",
+    )
```
victor-champonnois, line 76: The field is named `main_seller_id` but it points to a name, shouldn't it be called `main_seller_id_name` instead (would also be consistent with main_seller_id_product_code) ?

victor-champonnois (changes requested):

victor-champonnois:
Just some naming questions :)

beesdoo_product/models/beesdoo_product.py
remytms: You are right. But that's an odoo choice to put a res.partner into a field `name` of a supplierinfo object. Strange ideed, but cannot change that.

beesdoo_product/models/beesdoo_product.py
remytms: Because this field is a technical field, and it's not the main supplier that is selected but the first one. So top_supplier reflect more the reality.

beesdoo_product/models/beesdoo_product.py
remytms: supplierinfo object is a glue object between a partner (which is the seller) and a product.

remytms: Using `main_seller_id_name` will have no sens because we will expect this variable to contain a string that is the name of the supplier which is not. Because in `name` there is a res.partner. :)

remytms: I'm not sure to be clear. ^^

beesdoo_product/models/beesdoo_product.py
victor-champonnois: OK :+1:

beesdoo_product/models/beesdoo_product.py
victor-champonnois: OK, no it's clear :) Thanks for the explanation

victor-champonnois (approved):

Issue #247, opened 2021-07-13
RFC Moving Obeesdoo and Easy My Coop to OCA


robinkeunen:
Obeesdoo was mainly developed by the BEES coop volunteers and Coop IT Easy. Because of that, the implementation bears the BEES name and follows closely the processes at the BEES coop. Obeesdoo depends on Easy My Coop, also developed originally for the BEES coop. Both projects started in 2015.

Today, 11+ cooperative supermarket and grocery stores in Belgium, Spain, France, Germany and Switzerland are using Obeesdoo and 25+ cooperatives are using Easy My Coop. Additionally other IT cooperatives (coopdevs, cooperatic) are starting to use and contribute to the solution for their own customers.

At Coop IT Easy, we believe it's time to go further than just applying the AGPL license and to *really* make this project Open Source. By that we mean

- bring together a community around the code,
- make the code easier to contribute to
- **moving under the OCA umbrella**

## Why do it ?

- Raise the code quality to the OCA standards.
- Release the project ownership ;
- Become the reference repository for *Share Management* and *Cooperative Supermarket* with Odoo ;
- Make contributions easier for people outside Coop IT Easy ;
- Grow the project visibility and attract new customers and contributors ;
- Give a better understanding of the roadmap ;

### Required

1. Rename repositories and modules to more generic names
	- cf https://github.com/beescoop/Obeesdoo/issues/248 
2. Remove all references to Obeesdoo and Easy My Coop
	- in python code
	- and xml data
	- cf https://github.com/beescoop/Obeesdoo/issues/249 (archived)
3. Add demo data to demonstrate basic flows
	- cf https://github.com/beescoop/Obeesdoo/issues/250 (archived)
4. Cover main flows with unit tests
	- cf https://github.com/beescoop/Obeesdoo/issues/250 (archived)
5. Refactor code:
	- tasks already listed in Coop IT Easy's odoo
	- todo and fixme comments in the code 

#### Obeesdoo

1. Split modules by feature rather than by app
	- modules by feature can be expatriated to the other OCA repositories
	- modules by feature can be installed only for customer needing it
	- modules by feature are easier to maintain and migrate
2. centralize configurations : today configurations are set on the company, in res.settings, in module "configuration" menu and in other places. Gathering all configuration in one place would be more confortable and consistent for users (UX).

## How

The migration will be done **incrementally** in the order given in the previous section. Each improvement will bring its migration script relying on the *OCA Open Upgrade* Library.

[...]

PR #264, opened 2022-01-27
[WIP] Split beesdoo_product

carmenbianca:
WIP PR to split beesdoo_product into sale_suggested_price and product_label, although I suspect that it will be split up into more modules than two.

Some misc notes:

- Updated black because it would complain on me otherwise.

- Should this not be product_suggested_price?

- Make category product_margin default value configurable?

- `main_seller_id` was used in a view, but is no longer available to this module.

- There's a `views/point_of_sale_view.xml` file, but `point_of_sale` is not a  dependency. Should there be a module `sale_suggested_price_point_of_sale` or  some such that exists solely to add the view iff `point_of_sale` is installed?


carmenbianca:
Some notes on converting beesdoo_product into [product_labe;](https://github.com/grap/grap-odoo-business/tree/12.0/product_label). beesdoo_product contains considerably more functionality than suggested in #247. It contains:

- beesdoo.product.hazard:
  - Model with a name and type.
- product.template:
  - Labels (but different model structure from `grap/product_label`). Eco, local, fair, origin.
  - FDS (SDS in English?) label, which I believe is a hazard label, but it has a hazard label as well.
  - Setting main seller res.partner
  - `display_unit`, `default_reference_unit`, and `display_weight` to (I guess?) use a different uom internally and in outward views?
  - Calculating total prices with VAT. Also a deposit price.
  - Some interaction with printing labels?
  - Adding notes (comments)
  - Adding sell-before and best-before date
  - An ingredients field
  - Something to do with scales, scale categories, and scale labels?
  - Creating barcodes?
- beesdoo.scale.category
  - This entire model, which has a name and code
- beesdoo.product.label:
  - The labels described in product.template. name, type, and colour code.
- product.supplierinfo:
  - Price field added
- uom.category:
  - Type added? Not sure what this is. Unit, weight, time, distance, surface, volume, other.
- A bit of JavaScript code that interacts with the POS. Not sure what it does.
- Views:
  - Apart from adding fields to views…
  - Invoice policy is made invisible in `sale.product_template_form_view_invoice_policy`. Not sure why.
  - purchase_method, route_ids, and property_account_creditor_price_difference made invisible in `product.product_template_only_form_view`. Also not sure why.

I think that's fairly comprehensive. I'm not sure whether it's sensible to put all of that in product_label, but I'm not familiar enough with the codebase to make a good call on what to split where.

robinkeunen:
Cool, we'll make that call together when we get to this module. The main point of the exercise now is to  evaluate how long and complex the  migration script will (not) be so that we can plan the project with @polchampion

robinkeunen:
@carmenbianca can you  ?
- close this issue (we will open a new one when we reach this in the todo list)
- add the list of feature you identified to the module readme 
- push it on 12.0

PR #265, opened 2022-01-28
Fix missing computed shifts in /my/shift webpage

estevebadia:
Hello, this is Esteve from Manresa, Catalonia! Thanks sharing your work!

I think we found a little bug, hopefully this PR fixes it:

ISSUE:
Suppose a regular worker has two shifts already created. Then if this worker goes to /my/shift the page will not show their 3rd next shift. The page will show the first two (the created ones) and then the 4th and so on.

FIX:
The webpage /my/shift shows the future shifts for a regular worker. Some of these shifts are real (they are shift objects in the database) and the rest of them are "fake" for that page (no db records). 

The loop that creates these "computed" shifts takes the last real shift and adds a multiple of a period to compute the next shifts. So if the last shift is day D and the period between shifts is 28 days, the first fake shift needs to be at D+28, then at D+2*28 until we have a total of L (limit) shifts.

So I've modified the loop condition. It seems that it can also be fixed by taking not the last shift but the first one. However it seems better indeed to take the last shift. We could also just mutiply by (i-nb_subscribed_shifts+1) instead of by i at the end of the loop body, and not touching the loop condition.

Regards!

### DIFF
--- a/beesdoo_website_shift/controllers/main.py
+++ b/beesdoo_website_shift/controllers/main.py
@@ -430,7 +430,9 @@ class WebsiteShiftController(http.Controller):
                 .get_param("beesdoo_website_shift.shift_period")
             )
 
-            for i in range(1, regular_next_shift_limit - nb_subscribed_shifts + 1):
+            for i in range(
+                1, regular_next_shift_limit - nb_subscribed_shifts + 1
+            ):
                 # Create the fictive shift
                 shift = main_shift.new()
                 shift.name = main_shift.name
--- a/beesdoo_website_shift/controllers/main.py
+++ b/beesdoo_website_shift/controllers/main.py
@@ -430,7 +430,7 @@ class WebsiteShiftController(http.Controller):
                 .get_param("beesdoo_website_shift.shift_period")
             )
 
-            for i in range(nb_subscribed_shifts, regular_next_shift_limit):
+            for i in range(1, regular_next_shift_limit-nb_subscribed_shifts + 1):
                 # Create the fictive shift
                 shift = main_shift.new()
                 shift.name = main_shift.name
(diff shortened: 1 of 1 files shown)


robinkeunen:
Hi Esteve, thank you for your contribution ! I'll ping @remytms on this one, he know much more than I do about shifts.

robinkeunen:
By the way, did you take a look at this RFC https://github.com/beescoop/Obeesdoo/issues/247 ? Things will move a bit around here :-)

estevebadia:
> By the way, did you take a look at this RFC #247 ? Things will move a bit around here :-)

Exciting news! I'll keep an eye out for updates on that. It is an important step forward, but it will be a lot of work! Thanks for your work! And count on us for whatever you think we may help!

robinkeunen:
Thanks! We're still organising things out but the work will be planned out here: https://github.com/orgs/coopiteasy/projects/1/views/6

@remytms can you check this out ? :-)

robinkeunen:
Hi @estevebadia. I'm really sorry about the late review, Rémy has been overloaded recently. Can you please rebase your work on branch 12.0 which now contains unit tests ? I promise a swift review afterwards :-)

estevebadia:
Done! And don't worry fotr the timing. We don't need the changes to be merged quickly since we work on a forked version of your module set.

robinkeunen:
@estevebadia just a little lint issue :-)

MatteoLdx:
Hello ! Thanks a lot for this fix, it has been integrated to [this PR](https://github.com/beescoop/Obeesdoo/pull/331) that aims to fix another bug in this function.

estevebadia:
Perfect thank you! So closing this PR :)

PR #269, opened 2022-02-04
[REM] partner_firstname

huguesdk:
this module is an independent, simpler implementation than the oca module:
https://github.com/OCA/partner-contact/tree/9.0/partner_firstname

but it is the oca module that is used in production instead of this one, so
removing it to not shadow the oca one. it is still available in the
9.0-partner_firstname branch.


robinkeunen (approved):

robinkeunen:
@huguesdk can you merge or close ?

PR #273, opened 2022-02-14
[ADD] beesdoo_shift: next_shift_id on cooperative.status

robinkeunen:
- add a boolean : worker subscribed before alert date
- add next shift and next shift date to the worker status form

I needed to add `shift_shift_ids` to `res.partner` to trigger the computation. I hope it does not impact performance, maybe @tfrancoi knows.

- Question to @remytms : should I make my function depend on `today` field ?
- I could not write tests for `is_subscribed_to_shift` computation since the `next_alert_date" is computed in overriding modules.

[task](https://gestion.coopiteasy.be/web?#id=7720&view_type=form&model=project.task&action=479)

### DIFF
--- a/beesdoo_shift/models/cooperative_status.py
+++ b/beesdoo_shift/models/cooperative_status.py
@@ -1,5 +1,5 @@
 import logging
-from datetime import timedelta
+from datetime import datetime, timedelta
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError, ValidationError
@@ -65,7 +65,7 @@ class CooperativeStatus(models.Model):
         "are based on the current date",
         default=fields.Date.today,
     )
-    cooperator_id = fields.Many2one("res.partner")
+    cooperator_id = fields.Many2one("res.partner")  # todo rename to worker_id
     active = fields.Boolean(related="cooperator_id.active", store=True, index=True)
     info_session = fields.Boolean("Information Session ?")
     info_session_date = fields.Date("Information Session Date")
@@ -113,7 +113,23 @@ class CooperativeStatus(models.Model):
     irregular_absence_counter = fields.Integer()  # TODO unsubscribe when reach -2
     future_alert_date = fields.Date(compute="_compute_future_alert_date")
     next_countdown_date = fields.Date(compute="_compute_next_countdown_date")
-
+    next_shift_id = fields.Many2one(
+        comodel_name="beesdoo.shift.shift",
+        string="Earliest Open Shift",
+        compute="_compute_next_shift",
+        store=True,
+        help="Earliest open shift the worker is subscribed to.",
+    )
+    next_shift_date = fields.Datetime(
+        related="next_shift_id.start_time",
+        string="Next Shift Date",
+        store=True,
+    )
+    is_subscribed_to_shift = fields.Boolean(
+        string="Subscribed before Alert Date",
+        compute="_compute_next_shift",
+        store=True,
+    )
     temporary_exempt_reason_id = fields.Many2one(
         comodel_name="cooperative.exempt.reason",
         string="Temporary Exempt Reason",
@@ -379,6 +395,50 @@ class CooperativeStatus(models.Model):
         """
         for rec in self:
             rec.next_countdown_date = False
+
+    @api.multi
+    @api.depends(
+        "cooperator_id.shift_shift_ids",
+        "cooperator_id.shift_shift_ids.state",
+        "future_alert_date",
+        "today",
+    )
+    def _compute_next_shift(self):
+        # avoid searching for shift in loop
+        cooperator_ids = self.mapped("cooperator_id").ids
+        # rather take earliest "open" (confirmed) shift
+        # in order to take into account partners for which
+        # the counter was not incremented (happens when passing to "done")
+        next_shifts = self.env["beesdoo.shift.shift"].search(
+            [("state", "=", "open"), ("worker_id", "in", cooperator_ids)],
+            order="worker_id, start_time",
+        )
+
+        next_shift_by_worker = {}
+        for shift in next_shifts:
+            #  take first shift for each worker
+            next_shift_by_worker.setdefault(shift.worker_id.id, shift)
+
+        # what happens when no future shift ?
+        for rec in self:
+            rec.next_shift_id = next_shift_by_worker.get(rec.cooperator_id.id)
+
+            if not rec.next_shift_id:
+                rec.is_subscribed_to_shift = False
+                continue
+
+            if not rec.future_alert_date:
+                rec.is_subscribed_to_shift = True
+                continue
+
+            future_alert_date_time = datetime(
+                rec.future_alert_date.year,
+                rec.future_alert_date.month,
+                rec.future_alert_date.day,
+            )
+            rec.is_subscribed_to_shift = (
+                rec.next_shift_id.start_time < future_alert_date_time
+            )
 
     def _can_shop_status(self):
         """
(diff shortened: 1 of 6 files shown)


beesdoo_shift/models/cooperative_status.py
```diff
@@ -403,6 +422,38 @@ def _compute_next_countdown_date(self):
...
+                    rec.future_alert_date.month,
+                    rec.future_alert_date.day,
+                )
+                rec.is_subscribed_to_shift = (
+                    rec.next_shift_id.start_time < future_alert_date_time
+                )
+            if rec.next_shift_id and not rec.future_alert_date:
+                rec.is_subscribed_to_shift = True
+            else:
+                rec.is_subscribed_to_shift = False
```
victor-champonnois, line 427: won't this statement overide the statement after `if rec.next_shift_id and rec.future_alert_date` if `rec.next_shift_id` and `rec.future_alert_date` are true ?
Wondering if a structure like
```
 if rec.next_shift_id:
     if rec.future_alert_date:
     ...
     else:
     ...
else:
    ...
```
would be clearer ? (just wondering :) )

beesdoo_shift/models/cooperative_status.py
```diff
@@ -403,6 +422,38 @@ def _compute_next_countdown_date(self):
...
+    @api.multi
+    @api.depends(
+        "cooperator_id.shift_shift_ids",
+        "cooperator_id.shift_shift_ids.start_time",
+        "future_alert_date",
+    )
+    def _compute_next_shift(self):
+        now = fields.Datetime.now()
+        for rec in self:
+            rec.next_shift_id = self.env["beesdoo.shift.shift"].search(
```
tfrancoi: The search should be outside the loop, this kind of code is the root of performance issue
search worker_id in self.mapped('cooperator_id').ids and store the result in a dict 
You will send only one query for instead of one for each record.

victor-champonnois (commented):
FYI Read the code (more for learning than a proper review :) )

beesdoo_shift/models/cooperative_status.py
robinkeunen: Yes it would, I was missing a `elif`, good catch!

I restructured the code as you suggested.

beesdoo_shift/models/cooperative_status.py
robinkeunen: Ah, thanks. Something like this then ?

```py
    def _compute_next_shift(self):
        now = fields.Datetime.now()
        cooperator_ids = self.mapped("cooperator_id").ids
        # avoid searching for shift in loop
        next_shifts = self.env["beesdoo.shift.shift"].search(
            [("start_time", ">=", now), ("worker_id", "in", cooperator_ids)],
            order="worker_id, start_time",
        )

        next_shift_by_worker = {}
        for shift in next_shifts:
            #  take fist shift for each worker
            if shift.worker_id not in next_shift_by_worker:
                next_shift_by_worker[shift.worker_id] = shift

        for rec in self:
            rec.next_shift_id = next_shift_by_worker[rec.cooperator_id]
```

Thanks for the review

beesdoo_shift/models/cooperative_status.py
```diff
@@ -1,11 +1,13 @@
...
-from datetime import timedelta
+from datetime import datetime, timedelta
 
 from odoo import _, api, fields, models
 from odoo.exceptions import UserError, ValidationError
 
 _logger = logging.getLogger(__name__)
 
+# todo rename mentions to cooperatives
+
```
carmenbianca: If this still needs to be done in this PR: There's a TODO here that still needs doing :)

If this is more long-term: I think it's good practice to create an issue for this.

Same for other TODOs.

beesdoo_shift/models/cooperative_status.py
```diff
@@ -410,6 +429,52 @@ def _compute_next_countdown_date(self):
...
+        next_shifts = self.env["beesdoo.shift.shift"].search(
+            [("start_time", ">=", now), ("worker_id", "in", cooperator_ids)],
+            order="worker_id, start_time",
+        )
+
+        next_shift_by_worker = {}
+        for shift in next_shifts:
+            #  take fist shift for each worker
+            if shift.worker_id.id not in next_shift_by_worker:
+                next_shift_by_worker[shift.worker_id.id] = shift
```
carmenbianca:
```suggestion
            next_shift_by_worker.setdefault(shift.worker_id.id, shift)
```

beesdoo_shift/models/res_partner.py
```diff
@@ -66,6 +68,24 @@ class ResPartner(models.Model):
     subscribed_shift_ids = fields.Many2many(
         comodel_name="beesdoo.shift.template", readonly=True
     )
+    shift_shift_ids = fields.One2many(
+        comodel_name="beesdoo.shift.shift",
+        inverse_name="worker_id",
+        string="Shifts",
+        help="All the shifts the worker is subscribed to.",
+    )
```
carmenbianca, line 74: Should this not be computed? Where is this set?

carmenbianca (commented):
Some comments. Looks good overall. Tests are :ok_hand:

beesdoo_shift/models/res_partner.py
robinkeunen: One2many fields are computed by Odoo based on the Many2one fields in the co-model through the "inverse_name" field.

beesdoo_shift/models/cooperative_status.py
robinkeunen: It's a long term todo. We left a lot of long-term todo in the code but I agree with your suggestion. But since the code will be moved to `shift-planning` repository, we would need to migrate all issues as well. What do you think we should do ?
- create issue here then migrate all issues ?
- create issues in `shift-planning` directly although code does not exist ?
- leave the comments and create issues as we move the modules to `shift-planning` ?

beesdoo_shift/models/cooperative_status.py
robinkeunen: Although one liners are nice, I think we lose in readability with setdefault (I had to lookup the function doc).

beesdoo_shift/models/cooperative_status.py
carmenbianca: I think creating the issues in shift-planning directly is probably fine, even though the code is not there yet. I would probably create the issues, then edit the TODO comments to refer to the issues (full URL).

beesdoo_shift/models/cooperative_status.py
carmenbianca: I could go both ways. I've come to accept `setdefault` as a poorly named standard feature of Python dicts, and have just memorised it.

The advantage of `setdefault` to me is that—once you've memorised the awful name—you see immediately what it does. Whereas it took me a little while to understand what this for-loop was doing.

But there's no harm to me in doing it either way.

beesdoo_shift/models/cooperative_status.py
robinkeunen: I listed the todo in the general project task : https://github.com/coopiteasy/vertical-cooperative/issues/204#issue-1121003036

robinkeunen: and removed the todo

beesdoo_shift/models/cooperative_status.py
robinkeunen: Ok let's see if my brain gets used to it.

remytms (approved):
LGTM

PR #277, opened 2022-02-24
[ADD] beesdoo_pos_email_ticket: send ticket by email in pos

tfrancoi:
Not to be included in oca module but we can leave it in 12.0 branch of Obeesdoo


robinkeunen (approved):
Quick review since it won't be used on our infrastructure.

PR #281, opened 2022-02-24
[FIX] beesdoo_product: fix computed field total_with_vat

tfrancoi:
- Don't raise ValidationError in stored computed field it may crash at module installation
- Handle case were deposit is tax excluded, exclude them from the tax computation
  and include them at the end in the deposit field
- Error message is displayed on the product if the total can't be computed because of multiple tax strategies

Screenshot 2022-06-23 at 16 22 55

### DIFF
--- a/beesdoo_product/__manifest__.py
+++ b/beesdoo_product/__manifest__.py
@@ -15,7 +15,7 @@
     "author": "Beescoop - Cellule IT, Coop IT Easy SCRLfs",
     "website": "https://github.com/beescoop/Obeesdoo",
     "category": "Sales",
-    "version": "12.0.1.2.0",
+    "version": "12.0.1.3.0",
     "depends": [
         "product",
         "sale",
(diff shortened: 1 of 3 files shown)


tfrancoi:
Decide what we do in case of taxe included and excluded ?
- Let 0.0 and user has to guess
- Log the info : the user won't see it
- Display a nice message with a new computed : More work

robinkeunen:
Decision : 
- Ajouter un message à partir d'un champ calculé "ne peut pas être imprimé"

robinkeunen (changes requested):
Ajouter un message à partir d'un champ calculé "ne peut pas être imprimé".

polchampion:
Successfully tested - excluded-tax deposits are properly handled in SOs and on POS orders and tickets, even when other tax included.

robinkeunen (approved):

remytms (approved):
LGTM

PR #282, opened 2022-02-24
[IMP] beesdoo_product: make suggested price rounding configurable

tfrancoi:
Allow to configure the rounding of suggested price
choose the method and the precision of the rounding on the category

### DIFF
--- a/beesdoo_product/models/beesdoo_product.py
+++ b/beesdoo_product/models/beesdoo_product.py
@@ -330,6 +330,7 @@ class BeesdooProduct(models.Model):
             supplier = product._get_main_supplier_info()
             if supplier:
                 price = supplier.price
+                product_category = supplier.product_tmpl_id.categ_id
                 supplier_taxes = product.supplier_taxes_id.filtered(
                     lambda t: t.amount_type == "percent" and t.price_include
                 )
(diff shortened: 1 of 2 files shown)


beesdoo_product/models/beesdoo_product.py
```diff
@@ -451,6 +457,17 @@ def _check_margin(self):
             if product.profit_margin < 0.0:
                 raise UserError(_("Percentages for Profit Margin must >= 0."))
 
+    def _round(self, price):
+        self.ensure_one()
+        # Use default value 0.05 and "HALF_UP"
+        # in case someone erase the value on the category
+        # Keep default value on the field to make it explicit to the end user
```
robinkeunen, line 468: Can't we rather set the rounding method and precision to required ?

robinkeunen (approved):
Small comments. Tests would be nice, we'll do it when splitting the module and move it to the oca.

beesdoo_product/models/beesdoo_product.py
robinkeunen, line 468: Can't we rather set the rounding method and precision to required ?

robinkeunen (changes requested):

polchampion:
@robinkeunen Testé sur oufticoop-test et OK

robinkeunen (approved):

vdewulf:
Pol has documented this new feature here:
https://doc.it4socialeconomy.org/books/configurations/page/ajouter-sur-la-fiche-article-les-labels-prix-suggere-et-dangerosite

PR #285, opened 2022-02-24
[FIX] beesdoo_base: fix new eater

tfrancoi:
- is_customer is now call customer
- when an new eater is created the member card should be printed


robinkeunen (approved):

vdewulf:
Tested OK
on oufticoop.test.coopiteasy.be
I created a new Eater on an existing customer (worker and eater).
The newly created Eater has the boolean "Print BEES card?" set to True in the Member Card tab.

PR #286, opened 2022-02-24
[FIX] beesdoo_easy_my_coop: _cooperator_share_type return type

tfrancoi:
_cooperator_share_type should always return a record set
even if it's an empty one


robinkeunen (approved):
Not strictly necessary from the usage I find but cleaner indeed.

PR #288, opened 2022-02-24
[ADD] beesdoo_shift_welcome_screen: new module

tfrancoi:
Add a module for cooperator welcome screen

- Scan your member card
- See if you can shop
- See your counter and status

- See your eater
- See your next shift


beesdoo_shift_welcome_screen/models/welcome.py
```diff
@@ -0,0 +1,53 @@
...
+            barcode = '0' + barcode
+        if not barcode.startswith('042'):
+            self.message = 'invalid barcode'
+            return
+        #0 at the begining of the code bar seems not to be scanned
+        partner_ids = self.env['res.partner'].search([('barcode', '=', barcode)])
+        if not partner_ids:
+            self.message = "Member does not exist"
+        elif len(partner_ids) > 1:
+            self.message = "more then one member found with this barcode"
```
robinkeunen: Shouldn't all these messages be translated ?

beesdoo_shift_welcome_screen/models/welcome.py
```diff
@@ -0,0 +1,53 @@
...
+from odoo.exceptions import UserError
+
+class TaskType(models.TransientModel):
+    _name = 'beesdoo.shift.welcome'
+
+    _inherit = ['barcodes.barcode_events_mixin']
+
+    partner_id = fields.Many2one('res.partner', string="Cooperator")
+    message = fields.Html("Message")
+    # can_shop = fields.Boolean(related='partner_id.can_shop', readonly=True)
```
robinkeunen: remove this comment ?

beesdoo_shift_welcome_screen/models/welcome.py
```diff
@@ -0,0 +1,53 @@
...
+        }
+        html_res = self.env.ref("beesdoo_shift_welcome_screen.welcome_message").render(values)
+        self.message = html_res
+
+    def _get_next_shift(self, partner):
+        if not partner:
+            return self.env["beesdoo.shift.shift"]
+        now = datetime.now()
+        return self.env["beesdoo.shift.shift"].search([
+                ("start_time", ">", now.strftime("%Y-%m-%d %H:%M:%S")),
```
robinkeunen: Why not use a datetime type ? Des restes de la v9.

robinkeunen (changes requested):

vdewulf:
Welcome Screen tested on oufticoop.test.coopiteasy.be
New user created "welcome@oufticoop.be" with the right to "Shifts - Planning" (in order to see the Menu).

Connected as welcome user.
New Menu "Shift Welcome". Cf. screenshot
- TO TEST : scan a barcode.
I just entered a barcode manually in the Cooperator field, it does not work.
To be tested with a real scanner and a card.

- Manual selection
I selected a coop in the list (regular worker).
The status "Can shop" is displayed, his name, his counter and the next shifts (the next generated in the system it seems).
When the person in the list is an "eater", the name of the worker is displayed.
If the person has some eaters, their name is also displayed.


Conclusion : it works fine but we need to test the real "scan".
I edited the translation, cf. PO file for French BE (in CIE task because po file not supported in github)
welcome_screen_cannot_shop
welcome_screen_can_shop
welcome_screen_empty

polchampion:
Welcome Screen tested on bees.test.coopiteasy.be
Connected with admin (in order to see the Menu).
Tested with the scan for existing user.

FYI : Scan needs to be set up to be used aside of keyboard because it only works if the language is en_US. I changed my computer's language to en_US for it to work but then I can't use the keyboard in French (this is an unrelated issue).

Conclusion : it works.

Could be improved : when you select a cooperator in the list manually, then scan a barcode, it doesn't work (you need to click away first) : 
screenshot-bees test coopiteasy be-2022 03 29-15_26_51

polchampion:
@robinkeunen Tested (it works) and I left a suggestion above.

robinkeunen (approved):

PR #290, opened 2022-02-24
[FIX] beesdoo_shift: fix modularity

tfrancoi:
- Dynamic selection should be called with a lambda
- Clear signature for _get_irregular_worker_domain
- Modular list of field to watch for tracking history


robinkeunen (approved):

polchampion:
Tested on https://lenid.test.coopiteasy.be/

created a shift (create+duplicate): OK
added irregular worker to shift (online + backend): OK
validated attendance sheet: OK

polchampion (commented):
Tested with success on lenid-test

PR #291, opened 2022-02-24
Implement oca-ci via GitHub Actions

carmenbianca:


carmenbianca:
@robinkeunen The GitHub actions are not running on this PR. I suspect it's because the repo is configured to disallow non-members to muck about with GH actions.

Can you adjust those settings, or add me to beescoop? Or maybe a different solution?

In the meantime I'm opening a PR against my own fork for testing.

PR #292, opened 2022-02-24
[FIX] beesdoo_shift: forbid illegal hour

tfrancoi:
It was possible to encode shift that start at negative hour
or hour bigger then 24. Of course this lead to some
error at the moment we generate the planning for a given date

- Add a constraint on start_time and end_time
- Fix onchange on duration: duration was not computed properly
when the start or end time was 00:00

### DIFF
--- a/beesdoo_shift/models/planning.py
+++ b/beesdoo_shift/models/planning.py
@@ -214,10 +214,19 @@ class TaskTemplate(models.Model):
                     )
                 )
 
+    @api.constrains("start_time", "end_time")
+    def _check_real_hour(self):
+        for rec in self:
+            if not (-0.001 < rec.start_time < 24) or not (
+                -0.001 < rec.end_time < 24
+            ):
+                raise UserError(
+                    _("Start time and End time need to be between 0 and 23:59")
+                )
+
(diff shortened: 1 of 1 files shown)


beesdoo_shift/models/planning.py
```diff
@@ -214,10 +214,19 @@ def _nb_worker_max(self):
...
+            ):
+                raise UserError(
+                    _("Start time and End time need to be between 0 and 23:59")
+                )
+
     @api.onchange("start_time", "end_time")
     def _get_duration(self):
-        if self.start_time and self.end_time:
-            self.duration = self.end_time - self.start_time
+        self.duration = (self.end_time or 0.0) - (self.start_time or 0.0)
```
robinkeunen, line 229: I'd rather return False when `start_time` or `end_time` instead of these weird values.

robinkeunen:
```suggestion
        if self.start_time and self.end_time:
            self.duration = self.end_time - self.start_time
        else:
            self.duration = False

```

robinkeunen (changes requested):

PR #293, opened 2022-02-24
[MV] beesdoo_shift, beesdoo_worker_status: move constraint to specifi…

tfrancoi:
…c module

Constraint on shift type should be move to specific module for beescoop
workflow


robinkeunen (approved):

polchampion (approved):

tfrancoi:
@robinkeunen this one is now all green

polchampion:
Successfully tested: no error when using the coop_subscribe button

PR #297, opened 2022-02-24
Implement oca-ci via GitHub Actions

carmenbianca:
Both #291 and #296 wouldn't work for reasons I don't understand, but this should surely work.

edit: I was right :tada:


robinkeunen:
@carmenbianca I fixed the pre-commit stuff related to my merging frenzy of yesterday. I have not looked at the other test yet. If we can merge this soon, I'll rebase all the other (recently) open PR to get your CI EVEWYWERE

carmenbianca:
@robinkeunen We gotta calm down the ol' optimism :x We've hit a boatload of failing tests.

carmenbianca:
OK, all the failing tests are basically the same failure. `cooperative.status.working_mode` is not correctly set for the test users, and it doesn't like that.

Should that field be mandatory? Should there be a default?

carmenbianca:
Weird. Running the tests locally, they work :/ Gotta dig deeper.

carmenbianca:
Found the problem. oca-ci installs-then-tests the Obeesdoo modules in the same step. _For some reason_, this results in discrepant demo data.

Debugger inside of of test_change_worker_holiday_1 on the very first run:

```
>>> self.worker_regular_1.email
'fernand_peso@demode.net'
>>> self.worker_regular_1.working_mode
False
>>> self.worker_regular_1.cooperative_status_ids.working_mode
'regular'
```

On the second run:

```
>>> self.worker_regular_1.email
'fernand_peso@demode.net'
>>> self.worker_regular_1.working_mode
'regular'
>>> self.worker_regular_1.cooperative_status_ids.working_mode
'regular'
```

This is admittedly a little confusing… Is this known behaviour? Is this a bug that can be reported somewhere?

Do we have a preferred work-around? I can probably get oca-ci to install beesdoo_shift in a previous step.

carmenbianca:
In res.partner:

```python
    working_mode = fields.Selection(
        related="cooperative_status_ids.working_mode",
        readonly=True,
        store=True,
    )
```

Apparently this is a related field. This doesn't make a whole heap of sense.

carmenbianca:
4f7f3ee solved the issue quite a bit, though it's a silly workaround. Only 3 failing tests remaining.

carmenbianca:
Weird observation from some mucking about in odoo shell:

```python
>>> partner = self.env.ref(
            "beesdoo_shift.res_partner_worker_6_demo"
        )
>>> partner.working_mode
'regular'
>>> partner.cooperative_status_ids.working_mode
'regular'
>>> partner.working_mode = False
>>> partner.working_mode
False
>>> partner.cooperative_status_ids.working_mode
'regular'
```

:exploding_head: 

---

I'm able to reproduce the failing beesdoo_sheet_attendance test, and it seems to make sense that it fails. It finds `beesdoo_base.res_partner_cooperator_6_demo`, not to be confused with the almost identically named `beesdoo_shift.res_partner_worker_6_demo`. The former does not have any cooperative status, and therefore no defined working mode.

beesdoo_easy_my_coop/tests/test_res_partner.py
```diff
@@ -162,6 +162,9 @@ def test_compute_can_shop_share_a(self):
         Test that a cooperator can shop based on his share type.
         """
         coop1 = self.env.ref("beesdoo_base.res_partner_cooperator_1_demo")
+        coop1.cooperative_status_ids = self.env.ref(
+            "beesdoo_shift.cooperative_status_1_demo"
+        )
```
carmenbianca, line 171:
```suggestion
         coop1 = self.env.ref("beesdoo_shift.res_partner_worker_1_demo")
```

You can also attempt this.

carmenbianca:
It's alive! Thanks @coccinouille

beesdoo_easy_my_coop/tests/test_res_partner.py
victor-champonnois: Yes I tried this solution but it had complication because the `is_worker` field was overiden by beesdoo_easy_my_coop, so I needed to replace the attribute value.

carmenbianca: Ah, awesome. Can you add a fixup commit with a comment of what you just wrote? Some day it may be useful, especially if we refactor out the demo partner data from beesdoo_shift as we probably should.

coccinouille: You mean an empty fixup commit with just the comment ?

victor-champonnois: done :)

robinkeunen (approved):
🎉 can you squash the fixup commits before merging ?

robinkeunen:
@tfrancoi I merged this PR in order to get the ci on other PRs as well. It reintroduces the dependency of `beesdoo_shift` to `beesdoo_base` because `beesdoo_shift` still actually depended on member.card model defined in `beesdo_base`. I'll create a `member_card` module next week to remove that dependency again. So, in the meantime, no deployment in production recommended.

@huguesdk @remytms

tfrancoi:
@robinkeunen beesdoo_shift should not depends on beesdoo_base which part of the code user the member.card ? 
beesdoo_shift_welcome_screen should depends on it I agree but not beesdoo_shift

carmenbianca:
@tfrancoi 29b21ff fixed `test_attendance_sheet_barcode_scan` in beesdoo_shift_attendance. It required that the demo partners have member cards (which they didn't have). I see now that I could have defined those member cards in beesdoo_shift_attendance instead, but that would have been a slightly strange decision.

In any case, the reintroduced dependency will be removed very soon.

PR #299, opened 2022-02-24
[FIX] beesdoo_website_shift: fix public user access error

tfrancoi:
As public user you may face some access error
with beesdoo_shift website

the sudo are not a security issue since all those method return a boolean


robinkeunen (approved):

robinkeunen:
@remytms would like your point of view on this one, the fix is ok for me but do you know why there could be a bug in the first place ?

polchampion:
Tested - Couldn't reproduce the error.

Side find : As public user, you have access for a fraction of a second to the shifts.

https://user-images.githubusercontent.com/92738727/161264635-e8428fc5-b5aa-4c27-8191-1146a5830997.mp4

PR #300, opened 2022-02-24
[FIX] beesdoo_website_shift: ui fix

tfrancoi:
- Make the text "please subscribe" take the all width
- Add id for modularity


robinkeunen (approved):

polchampion (commented):
Tested with success on lenid-test

PR #301, opened 2022-02-24
[ADD] polln_shift: add specific polln shift

tfrancoi:
Module to handle specific workflow of polln


robinkeunen (approved):
merging because separate module + need it for other PR

polchampion (commented):
Tested on bees-test-polln-shift:
- irregular workers stay as they are: OK (note: working mode unchanged)
- impossible to register someone as a regular worker, though existing regular workers are not affected
- regular workers keep their slot, stay regular and can be resubscribed to new, single "worker" mode (= irregular): OK
- irregular workers can unsubscribe to shifts on the portal: NO - coulnd't find how
polln_unsubscribe_fail

- When subscribing a new coop to a shift, a checkbox enables automatic user creation: NO - coulnd't find how
screenshot-bees test coopiteasy be-2022 04 08-16_38_45

TO BE TESTED (reactivate crons):
- what happens to regular workers?
- test "worker" mechanism (no -2 penalty, alert at -2 (config), alert removed when back at 0)
- what implications to the normal flow ?



Test on oufticoop-test:
- Cannot install the module:
``` Odoo Server Error - Erreur de validation
Erreur lors de la validation de la contrainte

L'élément '<xpath expr="//p[@id='working_mode']">' ne peut être localisé dans la vue parente

Contexte de l'erreur :
Vue `Worker Status common`
[view_id: 1694, xml_id: n/a, model: n/a, parent_id: 1611] ```

PR #302, opened 2022-02-24
[IMP] beesdoo_shift: add name_get on task_template

tfrancoi:
this allow to use generated task template name
and not add all info of the template directly in the name

### DIFF
--- a/polln_shift/models/planning.py
+++ b/polln_shift/models/planning.py
@@ -41,6 +41,39 @@ class WizardSubscribe(models.TransientModel):
         return res
 
 
+class TaskTemplate(models.Model):
+    _inherit = "beesdoo.shift.template"
+
+    def name_get(self):
+        res = []
+        for rec in self:
+            name = "%s %s %s (%s-%s)" % (
+                rec.name,
+                rec.planning_id.name,
+                rec.day_nb_id.name,
+                float_to_time(rec.start_time),
+                float_to_time(rec.end_time)
+            )
+            res.append((rec.id, name))
+        return res
+
+    @api.model
+    def _name_search(self, name, args=None, operator='ilike', limit=100, name_get_uid=None):
+        domain = []
+        time_domain = []
+        FIELDS = ['planning_id', 'name', 'day_nb_id']
+        for n in name.split(" "):
+            if re.search(r"^\(\d{2}:\d{2}-\d{2}:\d{2}\)$", n):
+                start, end = n[1:-1].split('-')
+                start, end = time_to_float(start), time_to_float(end)
+                time_domain = [['&', ('start_time', '=', start), ('end_time', '=', end)]]
+            elif n.startswith('('): # Ignore missformed hour
+                continue
+            else:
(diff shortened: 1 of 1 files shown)


tfrancoi:
TODO: retarget on 12.0 once add-polln-shift is merged
Move this to beesdoo_shift

polln_shift/models/planning.py
```diff
@@ -41,6 +41,39 @@ def subscribe(self):
...
+            res.append((rec.id, name))
+        return res
+
+    @api.model
+    def _name_search(self, name, args=None, operator='ilike', limit=100, name_get_uid=None):
+        domain = []
+        time_domain = []
+        FIELDS = ['planning_id', 'name', 'day_nb_id']
+        for n in name.split(" "):
+            if re.search(r"^\(\d{2}:\d{2}-\d{2}:\d{2}\)$", n):
```
robinkeunen, line 66: could you provide an example(s) of matched pattern for readability ?
Also, nitpicking, shouldn't you compile the pattern out of the loop?

polln_shift/models/planning.py
```diff
@@ -41,6 +41,39 @@ def subscribe(self):
         return res
 
 
+class TaskTemplate(models.Model):
+    _inherit = "beesdoo.shift.template"
+
+    def name_get(self):
+        res = []
+        for rec in self:
```
robinkeunen, line 49: We try to use more explicit names in our loops, such as `for template in self:` or `for shift_template in self:`. Don't you use that at Odoo ?

robinkeunen (changes requested):
TODO: 
- will need a migration script.
- move to beesdoo_shift

polln_shift/models/planning.py
tfrancoi: > could you provide an example(s) of matched pattern for readability ? 
  Yes we are looking for the time (generated by the name get) for example 12:30-15h30 

> Also, nitpicking, shouldn't you compile the pattern out of the loop?
 From the perf point of view the number of iteration depends on the number of token in the search term so between 1 and 4. I don't think it really matter.

polln_shift/models/planning.py
tfrancoi: > Don't you use that at Odoo ?
There is no specific guidelines for that in odoo (you can grep for rec in self, for record in self) 
Personnaly when the record come from a search it makes sense to specify the object
when they come from self, you already know in which model you are, it's less important

polln_shift/models/planning.py
tfrancoi: https://stackoverflow.com/questions/452104/is-it-worth-using-pythons-re-compile

robinkeunen: Good to know, thanks!

polchampion:
Virg et Pol :  Ne pas merger, pas voulu, à mettre dans un module séparé si on veut vraiment implémenter cette fonctionnalié

polchampion (changes requested):
Ne pas merger, pas voulu, à mettre dans un module séparé si on veut vraiment implémenter cette fonctionnalié

PR #304, opened 2022-02-24
[IMP] beesdoo_shift: presence to shift can have variable value

tfrancoi:
Allow to have shift with a counter change value variable
and a float

### DIFF
--- a/polln_shift/models/planning.py
+++ b/polln_shift/models/planning.py
@@ -14,6 +14,12 @@ def time_to_float(t):
     return float(hour) + float(minute) / 60
 
 
+class TaskTemplate(models.Model):
+    _inherit = "beesdoo.shift.template"
+
+    shift_presence_value = fields.Float(default=1.0)
+
+
 class WizardSubscribe(models.TransientModel):
     _inherit = "beesdoo.shift.subscribe"
 
@@ -65,9 +71,9 @@ class Task(models.Model):
         data = {}
         status = self.worker_id.cooperative_status_ids[0]
         if new_state == "done":
-            data["sr"] = 1.0
+            data["sr"] = self.task_template_id.shift_presence_value or 1.0
         if new_state == "absent":
-            data["sr"] = -1.0
+            data["sr"] = -self.task_template_id.shift_presence_value or -1.0
         return data, status
 
     def _get_final_state(self):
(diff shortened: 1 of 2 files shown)


tfrancoi:
TODO: retarget on 12.0 once add-polln-shift is merged
Move this to beesdoo_shift
+ change sr to float (need a migration script)

robinkeunen (commented):
I'll leave this review to @remytms as he knows much more about counters than I do. Also, I don't think it's such a good idea to have a float for counters, can't you rather setup polln to have normal scripts increment by 2 and _half shifts_ increment by 1 ?

- Move this to beesdoo_shift
- change sr to float (need a migration script)

tfrancoi:
@robinkeunen I understand this change looks scary. We need more flexibiity then just normal presence and half and actually this change would require a bigger change than changing the field to float. 

But for the beescoop there is additionnal change require
because there is always the compensation counter
but in theory this require only to change this method and the compensation counter as well
https://github.com/beescoop/Obeesdoo/blob/12.0/beesdoo_worker_status/models/task.py#L55

If the beescoop does not this we can also either leave it in polln or move it to a sperate module (in both case we will need the trick of the counter stored field)

polchampion:
@tfrancoi If we're reading the code correctly, an absence to a shift of value 4 would decrement my counter by 4 points, correct?

polchampion:
Virg sonde le besoin en intercoop

polchampion:
@vdewulf Pour suivi

polchampion:
@vdewulf Un avis fonctionnel en passant : je vois surtout une complexité supplémentaire pour l'utilisateur shifteur et bureau des membres, au stade où on en est ça ne me semble pas bienvenu d'aller dans le sens d'un système plus complexe à utiliser.

vdewulf:
Réunion intercoop aujourd'hui : pas besoin d'ajouter de la complexité pour ajouter cette fonctionnalité qui n'est pas primordiale.
Donc à fermer selon moi
@robinkeunen

tfrancoi:
You can still merge this pr since it only impact polln_shift module it won't impact the others and in term of complexity, if the feature is not use it's just a field that can be hidden

tfrancoi:
You can still merge this pr since it only impact polln_shift module it won't impact the others and in term of complexity, if the feature is not use it's just a field that can be hidden

robinkeunen:
Ok, I'll rebase, fix precommit and merge.

robinkeunen (approved):
👍 approved since only in polln

PR #315, opened 2022-03-11
12.0 [ADD] solidarity shifts

MatteoLdx:
Add the ability to subscribe to a solidarity shift or to request one.


carmenbianca:
Suggestion to make this work again:

1. Satisfy pre-commit in #309.
2. Rebase on top of #309, but mark every solidarity-related entry as `edit`. Mark other entries as `drop`.
3. On every commit, resolve conflicts, then run `pre-commit` immediately afterwards, effectively removing the need for a "[FIX] Run pre-commit" commit.
4. Done.

victor-champonnois:
this PR breaks on the 12-test server, I remove it from cie-repositories

PR #320, opened 2022-03-15
[IMP] beesdoo_product: On changing purchase price, set sales price to suggested price

carmenbianca:
Requested by bees during komunigi meeting.

Task: https://gestion.coopiteasy.be/web#id=8283&view_type=form&model=project.task

Co-authored by @huguesdk

### DIFF
--- a/beesdoo_product/models/beesdoo_product.py
+++ b/beesdoo_product/models/beesdoo_product.py
@@ -296,51 +296,11 @@ class BeesdooProduct(models.Model):
         "categ_id.should_round_suggested_price",
     )
     def _compute_cost(self):
-        suggested_price_reference = (
-            self.env["ir.config_parameter"]
-            .sudo()
-            .get_param("beesdoo_product.suggested_price_reference")
-        )
         for product in self:
-            supplier = product._get_main_supplier_info()
-            if supplier:
-                price = supplier.price
-                product_category = supplier.product_tmpl_id.categ_id
-                supplier_taxes = product.supplier_taxes_id.filtered(
-                    lambda t: t.amount_type == "percent" and t.price_include
-                )
-                supplier_taxes_factor = 1 / (
-                    1 + sum(supplier_taxes.mapped("amount")) / 100
-                )
-                sale_taxes = product.taxes_id.filtered(
-                    lambda t: t.amount_type == "percent" and t.price_include
-                )
-                sale_taxes_factor = 1 + sum(sale_taxes.mapped("amount")) / 100
-                profit_margin_supplier = supplier.name.profit_margin
-                profit_margin_product_category = product_category.profit_margin
-                profit_margin = profit_margin_supplier or profit_margin_product_category
-                profit_margin_factor = (
-                    1 / (1 - profit_margin / 100)
-                    if suggested_price_reference == "sale_price"
-                    else (1 + profit_margin / 100)
-                )
-
-                # price of purchase is given for uom_po_id
-                #   suggested *sale* price must be adapted to uom_id
-                uom_factor = product.uom_po_id.factor / product.uom_id.factor
-
-                product.suggested_price = (
-                    price
-                    * uom_factor
-                    * supplier_taxes_factor
-                    * sale_taxes_factor
-                    * profit_margin_factor
-                )
-
-                if product_category.should_round_suggested_price:
-                    product.suggested_price = product_category._round(
-                        product.suggested_price
-                    )
+            try:
+                product.suggested_price = product.calculate_suggested_price()
+            except ValueError:
+                pass
 
     @api.multi
     @api.depends("seller_ids")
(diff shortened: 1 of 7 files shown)


beesdoo_product/models/beesdoo_product.py
```diff
@@ -390,6 +348,83 @@ def _inverse_purchase_price(self):
...
+        suggested_price_reference = (
+            self.env["ir.config_parameter"]
+            .sudo()
+            .get_param("beesdoo_product.suggested_price_reference")
+        )
+        supplier = self._get_main_supplier_info()
+        if not supplier:
+            raise ValueError(
+                _("No supplier found for product {}").format(self.id)
+            )
```
victor-champonnois: Should it really be considered an error if this error is catched and passed everytime the function is called ?

beesdoo_product/models/beesdoo_product.py
```diff
@@ -348,6 +358,13 @@ def _inverse_purchase_price(self):
                     _("No Vendor defined for product '%s'") % product.name
                 )
 
+    @api.multi
+    @api.depends("purchase_price", "seller_ids")
+    def _compute_purchase_price_write_date(self):
+        for product in self:
+            supplierinfo = product._get_main_supplier_info()
```
victor-champonnois, line 340: `supplier_info` looks better maybe ? (really really nitpicking here :smile: )

victor-champonnois (commented):
Read the code and it looks great :smile:

beesdoo_product/models/beesdoo_product.py
carmenbianca: Exceptions in Python are a little special. They're more frequently used as a program flow tool. At the end of every iteration, a `StopIteration` exception is thrown and caught by the interpreter to break a for-loop.

In this case, I need to convey to consumers of this function that no answer could be found. I could return None, but Python developers generally prefer catching an exception over checking against None.

beesdoo_product/models/beesdoo_product.py
carmenbianca: The model is called `product.supplierinfo` :)

victor-champonnois: ah ok, perfect then !

beesdoo_product/models/beesdoo_product.py
victor-champonnois: OK, thanks for the explanations :)

victor-champonnois (approved):

carmenbianca:
Merci @coccinouille !

beesdoo_product/models/beesdoo_product.py
```diff
@@ -363,6 +333,91 @@ def _inverse_purchase_price(self):
...
+                # 'calculate_suggested_price' twice, but that's just how it is.
+                suggested_price = self.calculate_suggested_price(price=purchase_price)
+
+                # Important note: The list price is _only_ changed here if the
+                # `purchase_price` field of the product is changed. If the
+                # `price` field of the supplierinfo changes, the list price here
+                # is NOT automatically affected.
+                self.adapt_list_price(vals, suggested_price=suggested_price)
+            except ValueError:
+                pass
```
huguesdk: the rationale behind `adapt_list_price()` was to avoid duplicating the logic (that determines where the new price is coming from) in the wizard and in `write()`. as it is, the function is not useful anymore. the problem comes from the fact that the wizard can use `product.suggested_price`, while in `write()` it’s not possible because the property is not yet up to date. could we maybe have something like this?
```suggestion
            # force update of purchase_price (which actually modifies the
            # suppliers’ price) to ensure that the next computations that
            # depend on it are correct. there are 3 things to do:
            # 1. set value in cache
            with self.env.do_in_draft():
                self.purchase_price = purchase_price
            # 2. call inverse compute method
            self._inverse_purchase_price()
            # 3. remove the value from vals, to avoid the process to happen a
            #    second time
            del vals["purchase_price"]

            self.adapt_list_price(vals)
```

this allows to keep the code as it was before: no need to move `suggested_price` computation in another method, and the relationship between `list_price` and `suggested_price` is only in `adapt_list_price()`. the downside is that it’s probably not very clean to modify fields like this, and that `_inverse_purchase_price()` is called explicitly in `write()` instead of using a generic method (but maybe there is a way?).

beesdoo_product/views/beesdoo_product.xml
```diff
@@ -110,13 +110,15 @@
...
                 <field name="purchase_price" />
+                <field name="purchase_price_write_date" />
                 <field
                     name="uom_po_id"
                     readonly="1"
                     options="{'no_open': True, 'no_create': True, 'no_create_edit':True }"
                 />
                 <field name="suggested_price" readonly="1" />
                 <field name="list_price" readonly="1" />
+                <field name="list_price_write_date" />
```
huguesdk: these new fields take a lot of space (width) in the list. shorter names would be useful. since they are next to their corresponding prices, maybe simply “Last Update” would be enough (set with the `string` property directly here in xml)?

also, i don’t think displaying seconds is useful (and makes it wider), but i don’t know whether this can be easily changed.

huguesdk (changes requested):

beesdoo_product/models/beesdoo_product.py
carmenbianca: I'm not entirely sure this works, and if it does work, I think it might break a (possibly implicit) rule:

1. `self.purchase_price` is set in draft mode and written to a cache. How long does this cache live? When is that cache emptied? Fortunately it's not a stored field, but if it were, this would assuredly break.
2. `self._inverse_purchase_price()` writes before `write()` is done. I'm not sure that sort of non-atomic operation is allowed. If `write()` is in any way interrupted, you get an inconsistent cache/database.
3. If `self._inverse_purchase_price()` writes something to the main supplier, then isn't `self.seller_ids` changed, and isn't `self._compute_cost()` going to trigger?

This is a stupid problem :/

beesdoo_product/views/beesdoo_product.xml
carmenbianca: Tried in 77c6919. I was unable to remove the seconds.

Old (before PR):

Capture d’écran de 2022-03-17 10-08-34

New (after 77c6919):

Capture d’écran de 2022-03-17 10-05-10

beesdoo_product/models/beesdoo_product.py
carmenbianca: Tested this and it works. Haven't found a way to break it. I'll push this as commit. If it doesn't work as desired, we can revert that specific commit.

huguesdk (approved):

robinkeunen:
@carmenbianca i think this can me merged so that @victor-champonnois can add it to https://github.com/beescoop/Obeesdoo/pull/393

victor-champonnois:
@carmenbianca can this be merged ? I need to merge it to my beesdoo_product split.

carmenbianca:
@victor-champonnois good news! we're merging this!

beesdoo_product/models/beesdoo_product.py
```diff
@@ -363,6 +333,98 @@ def _inverse_purchase_price(self):
...
+    @api.depends("purchase_price", "seller_ids")
+    def _compute_purchase_price_write_date(self):
+        for product in self:
+            supplierinfo = product._get_main_supplier_info()
+            product.purchase_price_write_date = supplierinfo.price_write_date
+
+    def write(self, vals):
+        if self.env["ir.config_parameter"].get_param(
+            "beesdoo_product.auto_write_suggested_price"
+        ):
```
MatteoLdx, line 346: Isn't a sudo() missing here ?

PR #324, opened 2022-03-18
[FIX] use max_nb_eater defined in the share

victor-champonnois:
## Description
Use the max number of eater defined in the share rather than the fixed amount

## Odoo task (if applicable)
https://gestion.coopiteasy.be/web#id=8305&view_type=form&model=project.task&action=475&active_id=224

### DIFF
--- a/beesdoo_pos/models/beesdoo_pos.py
+++ b/beesdoo_pos/models/beesdoo_pos.py
@@ -7,16 +7,8 @@ class BeescoopPosPartner(models.Model):
     @api.multi
     def get_eater(self):
         self.ensure_one()
-        eaters = [False] * self._get_max_nb_eater_allowed()
-        for i, eater in enumerate(self.child_eater_ids):
-            eaters[i] = eater.name
+        eaters = []
+        # todo check for max eater
+        for eater in self.child_eater_ids:
+            eaters.append(eater.name)
         return eaters
-
-    def _get_max_nb_eater_allowed(self):
-        if self.eater == "eater":
-            max_nb_eater_allowed = (
-                self.parent_eater_id._cooperator_share_type().max_nb_eater_allowed
-            )
-        else:
-            max_nb_eater_allowed = self._cooperator_share_type().max_nb_eater_allowed
-        return max_nb_eater_allowed
--- a/beesdoo_pos/models/beesdoo_pos.py
+++ b/beesdoo_pos/models/beesdoo_pos.py
@@ -6,6 +6,7 @@ class BeescoopPosPartner(models.Model):
 
     @api.multi
     def get_eater(self):
+        self.ensure_one()
         eaters = [False] * self._get_max_nb_eater_allowed()
         for i, eater in enumerate(self.child_eater_ids):
             eaters[i] = eater.name
@@ -17,7 +18,5 @@ class BeescoopPosPartner(models.Model):
                 self.parent_eater_id._cooperator_share_type().max_nb_eater_allowed
             )
         else:
-            max_nb_eater_allowed = (
-                self._cooperator_share_type().max_nb_eater_allowed
-            )
+            max_nb_eater_allowed = self._cooperator_share_type().max_nb_eater_allowed
         return max_nb_eater_allowed
--- a/beesdoo_pos/models/beesdoo_pos.py
+++ b/beesdoo_pos/models/beesdoo_pos.py
@@ -7,8 +7,5 @@ class BeescoopPosPartner(models.Model):
     @api.multi
     def get_eater(self):
         self.ensure_one()
-        eaters = []
         # todo check for max eater
-        for eater in self.child_eater_ids:
-            eaters.append(eater.name)
-        return eaters
+        return [eater.name for eater in self.child_eater_ids]
--- a/beesdoo_pos/models/beesdoo_pos.py
+++ b/beesdoo_pos/models/beesdoo_pos.py
@@ -4,13 +4,20 @@ from odoo import api, models
 class BeescoopPosPartner(models.Model):
     _inherit = "res.partner"
 
-    def _get_eater(self):
-        eaters = [False, False, False]
-        for i, eater in enumerate(self.child_eater_ids):
-            eaters[i] = eater.name
-        return tuple(eaters)
-
     @api.multi
     def get_eater(self):
-        eater1, eater2, eater3 = self._get_eater()
-        return eater1, eater2, eater3
+        eaters = [False] * self._get_max_nb_eater_allowed()
+        for i, eater in enumerate(self.child_eater_ids):
+            eaters[i] = eater.name
+        return eaters
+
+    def _get_max_nb_eater_allowed(self):
+        if self.eater == "eater":
+            max_nb_eater_allowed = (
+                self.parent_eater_id._cooperator_share_type().max_nb_eater_allowed
+            )
+        else:
+            max_nb_eater_allowed = (
+                self._cooperator_share_type().max_nb_eater_allowed
+            )
+        return max_nb_eater_allowed
--- a/beesdoo_pos/static/src/js/beesdoo.js
+++ b/beesdoo_pos/static/src/js/beesdoo.js
@@ -38,7 +38,7 @@ odoo.define("beesdoo_pos.screens", function (require) {
                     timeout: 1000,
                 }
             )
-                .then(function(result) {
+                .then(function (result) {
                     for (let i = 0; i < result.length; i++) {
                         set_customer_info.call(
                             self,
--- a/beesdoo_pos/static/src/js/beesdoo.js
+++ b/beesdoo_pos/static/src/js/beesdoo.js
@@ -38,25 +38,15 @@ odoo.define("beesdoo_pos.screens", function (require) {
                     timeout: 1000,
                 }
             )
-                .then(function (result) {
-                    set_customer_info.call(
-                        self,
-                        ".customer-delegate1",
-                        result[0],
-                        "Eater 1: "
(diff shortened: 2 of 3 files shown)


beesdoo_pos/models/beesdoo_pos.py
```diff
@@ -4,13 +4,20 @@
...
 
-    def _get_eater(self):
-        eaters = [False, False, False]
+    @api.multi
+    def get_eater(self):
+        eaters = [False] * self._get_max_nb_eater_allowed()
         for i, eater in enumerate(self.child_eater_ids):
             eaters[i] = eater.name
-        return tuple(eaters)
+        return eaters
```
robinkeunen: If you're adding `@api.multi`, you should either make sure your function works with a recordset of any length or add `self.ensure_one` to the beginning of the function.

robinkeunen (changes requested):

beesdoo_pos/models/beesdoo_pos.py
victor-champonnois: fixed, thx !

victor-champonnois:
@robinkeunen Do you think the JS part is OK ? I did it a bit wildly !

robinkeunen (approved):
I can't really comment on JS quality : I don't know it. The logic seems sound :-)

beesdoo_pos/models/beesdoo_pos.py
```diff
@@ -4,13 +4,19 @@
...
-    def get_eater(self):
-        eater1, eater2, eater3 = self._get_eater()
-        return eater1, eater2, eater3
+    def _get_max_nb_eater_allowed(self):
+        if self.eater == "eater":
+            max_nb_eater_allowed = (
+                self.parent_eater_id._cooperator_share_type().max_nb_eater_allowed
+            )
+        else:
+            max_nb_eater_allowed = self._cooperator_share_type().max_nb_eater_allowed
```
huguesdk: a comment would be nice. i had to go look into `beesdoo_base` to understand what the `else` case is. something like:
```python
# this partner is an eater, not a cooperator, so we need to access the cooperator partner to which it is linked to to access the share type.
```

beesdoo_pos/models/beesdoo_pos.py
```diff
@@ -4,13 +4,19 @@
...
-    def _get_eater(self):
-        eaters = [False, False, False]
+    @api.multi
+    def get_eater(self):
+        self.ensure_one()
+        eaters = [False] * self._get_max_nb_eater_allowed()
         for i, eater in enumerate(self.child_eater_ids):
             eaters[i] = eater.name
-        return tuple(eaters)
+        return eaters
```
huguesdk: why was it a tuple before and not anymore?

beesdoo_pos/static/src/js/beesdoo.js
```diff
@@ -39,24 +39,14 @@ odoo.define("beesdoo_pos.screens", function (require) {
...
-                        ".customer-delegate3",
-                        result[2],
-                        "Eater 3: "
-                    );
+                    for (let i = 0; i < result.length; i++) {
+                        set_customer_info.call(
+                            self,
+                            ".customer-delegate".concat(i + 1),
+                            result[i],
+                            "Eater".concat(i + 1, ": ")
```
huguesdk: there is a space missing after `Eater` (also on line 89).

huguesdk (changes requested):
the js part looks good to me, but [xml template](https://github.com/beescoop/Obeesdoo/blob/12.0-use-max_nb_eater-defined-in-the-share/beesdoo_pos/static/src/xml/templates.xml) has 3 `customer-delegate`s hard-coded (several times), so it won’t work if the number is higher than 3. the template should use a loop too.

beesdoo_pos/models/beesdoo_pos.py
```diff
@@ -4,13 +4,19 @@
...
     _inherit = "res.partner"
 
-    def _get_eater(self):
-        eaters = [False, False, False]
+    @api.multi
+    def get_eater(self):
+        self.ensure_one()
+        eaters = [False] * self._get_max_nb_eater_allowed()
         for i, eater in enumerate(self.child_eater_ids):
             eaters[i] = eater.name
```
huguesdk: a comment i wrote but forgot to add appparently:

another way to write this:
```python
eaters = [eater.name for eater in self.child_eater_ids)
# pad with False
eaters.extend([False] * (self._get_max_nb_eater_allowed() - len(eaters))
```
not sure which one is more readable, though. 😅

beesdoo_pos/static/src/js/beesdoo.js
huguesdk: also, does this work with translations?

huguesdk (commented):
since the number of eaters is variable, isn’t there a way to have a dynamic list of elements in a template without having to fiddle with number suffixes to identify them?

beesdoo_pos/models/beesdoo_pos.py
victor-champonnois: done thanks :)

victor-champonnois:
> since the number of eaters is variable, isn’t there a way to have a dynamic list of elements in a template without having to fiddle with number suffixes to identify them?

@huguesdk that's a very good idea! I refactored the whole thing, which also corrected for some issues you mentionned. 
- now get_eaters returns a list, since the number of eater is variable (the list is mutable so it seems like a better option)
- I fixed the other templates
-  I removed the "Eater" strings, which indeed where not translated. I'll ask the analyst if it's ok to remove them.

I noticed that the check for the max number of eaters is done [here](https://github.com/beescoop/Obeesdoo/blob/12.0/beesdoo_base/models/partner.py#L62), and it should also be refactored. I'll see with the analyst first.

beesdoo_pos/models/beesdoo_pos.py
```diff
@@ -4,13 +4,11 @@
...
     @api.multi
     def get_eater(self):
-        eater1, eater2, eater3 = self._get_eater()
-        return eater1, eater2, eater3
+        self.ensure_one()
+        eaters = []
+        # todo check for max eater
+        for eater in self.child_eater_ids:
+            eaters.append(eater.name)
+        return eaters
```
huguesdk: a shorter way of writing this:

```suggestion
        return [eater.name for eater in self.child_eater_ids]
```

beesdoo_pos/static/src/js/beesdoo.js
```diff
@@ -5,14 +5,11 @@ odoo.define("beesdoo_pos.screens", function (require) {
...
 
-    var set_customer_info = function (el_class, value, prefix) {
-        var el = this.$(el_class);
-        el.empty();
-        if (prefix && value) {
-            value = prefix + value;
-        }
+    var set_customer_info = function (parent_class, value) {
+        var parent = this.$(parent_class);
+        var info_div = $("<div></div>").text(value);
```
huguesdk: this seems like a strange way to create an element. i don’t know what the best way is, but here is another way:
```suggestion
        var info_div = document.createElement("div");
        info_div.textContent = value;
```
and then no need for the `[0]` below.

beesdoo_pos/static/src/js/beesdoo.js
```diff
@@ -91,23 +77,8 @@ odoo.define("beesdoo_pos.screens", function (require) {
...
-                        result[1],
-                        "Eater 2: "
-                    );
-                    set_customer_info.call(
-                        self,
-                        ".customer-delegate3",
-                        result[2],
-                        "Eater 3: "
+                    result.forEach((element) =>
+                        set_customer_info.call(self, ".customer-delegates", element)
```
huguesdk: i wouldn’t call the variable `element` because it implies that it is an html element, while it is the name of person (as a string), if i understand correctly. `customer_name`, maybe?

didn’t you lost the “Eater <n>: ” prefix? is this wanted?

huguesdk (changes requested):
great, much better! generic code, clearer and shorter. congrats!

beesdoo_pos/models/beesdoo_pos.py
victor-champonnois: Nice thanks :)

beesdoo_pos/static/src/js/beesdoo.js
victor-champonnois: - OK, I went for `client_name` instead, to be consistent with the `get_client` method.

- Yes I lost the "Eater 1:" prefix, I am waiting for the analyst's answer to see if it's OK or what would be a better output.

beesdoo_pos/static/src/js/beesdoo.js
victor-champonnois: Thx, I've read that the way I did this is the standard way with jquery ([see here ](https://www.w3schools.com/jquery/jquery_dom_add.asp)for instance), but your way is cleaner :)

victor-champonnois:
@huguesdk Thx for the re-review :) I accounted for all your comment in the code.

beesdoo_pos/static/src/js/beesdoo.js
huguesdk: thanks for the information. the jquery way is ok for me (although there is no `[0]` in the example). in any case, a comment explaining what is being done would be nice to have, since it’s not obvious. something like: `create a new div containing value`.

victor-champonnois: Sorry I was not clear, I used your way in the new fixup commit :)

huguesdk (approved):
lgtm 👍

PR #331, opened 2022-04-04
[FIX] b_website_shift: fix the future shifts display on "My shifts" webpage

MatteoLdx:
### First issue
ISSUE:
When calculating next shifts to display on webpage "My shifts" for regular workers, the shift used as a reference was the last generated shift of the worker. This caused a problem if the last shift was an "unusual" shift (for exemple, if the worker has been allowed to change the time of his/her shift one day because he/she couldn't attend the usual one).

FIX:
This PR tries to fix this problem by always taking as reference the last shift that corresponds to the template linked to the worker subscription.

### Second issue (credits to @estevebadia)
ISSUE:
Suppose a regular worker has two shifts already created. Then if this worker goes to /my/shift the page will not show their 3rd next shift. The page will show the first two (the created ones) and then the 4th and so on.

FIX:
The webpage /my/shift shows the future shifts for a regular worker. Some of these shifts are real (they are shift objects in the database) and the rest of them are "fake" for that page (no db records).
The loop that creates these "computed" shifts takes the last real shift and adds a multiple of a period to compute the next shifts. So if the last shift is day D and the period between shifts is 28 days, the first fake shift needs to be at D+28, then at D+2*28 until we have a total of L (limit) shifts.
So I've modified the loop condition. It seems that it can also be fixed by taking not the last shift but the first one. However it seems better indeed to take the last shift. We could also just mutiply by (i-nb_subscribed_shifts+1) instead of by i at the end of the loop body, and not touching the loop condition.

## Checklist before approval

### DIFF
--- a/beesdoo_website_shift/controllers/main.py
+++ b/beesdoo_website_shift/controllers/main.py
@@ -389,28 +389,24 @@ class WebsiteShiftController(http.Controller):
         # according to the regular_next_shift_limit
         if self.is_user_regular():
             # Compute main shift
-            nb_subscribed_shifts = len(subscribed_shifts)
-            if nb_subscribed_shifts > 0:
-                main_shift = subscribed_shifts[-1]
-            else:
-                task_template = (
-                    request.env["beesdoo.shift.template"]
-                    .sudo()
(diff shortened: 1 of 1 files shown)


remytms (approved):
LGTM.

PR #376, opened 2022-04-28
[FIX] beesdoo_pos: Fix corner-case where get_client() returns a different value

carmenbianca:
## Description

Because of asynchronicity fun, get_client() would sometimes return null,
and thus raise an error.

Related to https://github.com/coopiteasy/addons/pull/222, which _somehow_ maybe possibly triggered this behaviour.

## Odoo task (if applicable)

Bug was found here: https://gestion.coopiteasy.be/web#id=4393&view_type=form&model=project.task

This task probably more accurately encapsulates this bug: https://gestion.coopiteasy.be/web#id=8305&view_type=form&model=project.task&action=475&active_id=224

## Checklist before approval

### DIFF
--- a/beesdoo_pos/static/src/js/beesdoo.js
+++ b/beesdoo_pos/static/src/js/beesdoo.js
@@ -19,15 +19,15 @@ odoo.define("beesdoo_pos.screens", function (require) {
             var self = this;
             var loaded = new $.Deferred();
             this._super();
-            if (!this.pos.get_client()) {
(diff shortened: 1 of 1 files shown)


carmenbianca:
I'm starting to suspect that #324 is broken. I'll write an Odoo task about this to elaborate :/

edit: https://gestion.coopiteasy.be/web#id=8493&view_type=form&model=project.task&action=475&active_id=264

PR #378, opened 2022-04-29
12.0 fix pos client name list

victor-champonnois:
Bug fix related to [this task](https://gestion.coopiteasy.be/web#id=8305&view_type=form&model=project.task&action=475&active_id=224)

Fix display of customers  and eaters name displaying in the payment screen.
- don't create new div but empty and append existing div
- remove customer name from `paymentmethods-container`
- add @carmenbianca 's fix from https://github.com/beescoop/Obeesdoo/pull/376

### DIFF
--- a/beesdoo_pos/static/src/js/beesdoo.js
+++ b/beesdoo_pos/static/src/js/beesdoo.js
@@ -5,12 +5,10 @@ odoo.define("beesdoo_pos.screens", function (require) {
 
     models.load_fields("res.partner", "is_company");
 
-    var set_customer_info = function (el, value, prefix) {
-        if (prefix && value) {
-            value = prefix + value;
-        }
+    var set_customer_info = function (el, value) {
+        el.empty();
         if (value) {
-            el.append(value + "<br />");
+            el.append(value);
         }
     };
     screens.ActionpadWidget.include({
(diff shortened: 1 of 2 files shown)


beesdoo_pos/static/src/js/beesdoo.js
```diff
@@ -5,29 +5,30 @@ odoo.define("beesdoo_pos.screens", function (require) {
...
+    var set_customer_info = function (el_class, value, prefix) {
+        var el = this.$(el_class);
+        el.empty();
+        if (prefix && value) {
+            value = prefix + value;
+        }
         if (value) {
-            parent.append(info_div);
+            el.append(value + "<br />");
         }
```
carmenbianca:
```suggestion
        if (prefix && value) {
            value = prefix + value;
        }
        this.$(el_class).text(value)
```

something like this? no need to empty+append when you can just set the text immediately.

carmenbianca (changes requested):

beesdoo_pos/static/src/js/beesdoo.js
victor-champonnois: much more concise, great :+1: !

victor-champonnois: Actually no the append is needed in case of multiple eaters, otherwise it will just show one. it is the empty that is problematic

victor-champonnois: OK so I moved the `empty` out of the function to trigger it only once

carmenbianca: @victor-champonnois I wouldn't do it like that. I would just call the function with a value of a `<br />` separated string of names. So instead of calling `set_customer_info` each time for every eater, just call it once.

victor-champonnois: Fixed, thanks for the proposition :)

victor-champonnois:
@carmenbianca Is it OK to merge ?

carmenbianca:
@victor-champonnois I'm going to do a little bit of rebasing on this branch.

carmenbianca:
Merged this locally.

PR #379, opened 2022-04-29
[IMP] b_website_shift_swap: display shifts as a calendar to swap or offer solidarity

MatteoLdx:
## Description

Change the display of the shifts on the pages where a selection is needed, from a list to a calendar.

## Checklist before approval


remytms (changes requested):
Try to use the original `build_shift_grid` instead of a custom one. :)

PR #388, opened 2022-05-19
[FIX] beesdoo_worker_status: Try not to apply penalty twice in a row

carmenbianca:
## Odoo task (if applicable)

https://gestion.coopiteasy.be/web#view_type=form&model=project.task&id=6332&active_id=6332&menu_id=

## Checklist before approval


huguesdk (commented):
i don’t like so much naming a field `skip_something` because it’s not that clear. i think it would be better to use a name that indicates a state, like `is_in_penalty` (not sure about the `is_` prefix, though, what is usual in odoo?). in addition to being more clear, it allows to change the logic later on, as it decouples the name from the behavior.

carmenbianca:
@huguesdk good observation. I'll do a little research and add a patch to adjust the field name.

huguesdk (approved):
👍

remytms (approved):
Nice

PR #416, opened 2022-06-20
12.0 [FIX] beesdoo_shift_swap

MatteoLdx:
## Description

Bugfixes and improvements for code of PR #309 

## Checklist before approval


beesdoo_shift/models/task.py
```diff
@@ -39,6 +39,9 @@ def _get_color_mapping(self, state):
     def _get_final_state(self):
         return ["done", "absent", "excused"]
 
+    def get_absent_state(self):
+        return ["absent"]
+
```
remytms, line 45: You may want to add `@api.model` to this function, as it does not depends on the data of self.

beesdoo_worker_status_shift_swap/models/task.py
```diff
@@ -14,11 +14,7 @@ def _get_counter_date_state_change(self, new_state):
             and new_state in ["done", "absent_0"]
             and self.solidarity_offer_ids[0]
         ):
-            data["sr"] = 0
```
remytms, line 17: Are you sure to delete this line ?

remytms (commented):
Check my comments to be sure. :)

remytms:
@MatteoLdx I think that changing `SCRLfs` to `SC` is a little bit to early, because the CI is not ready yet to check that correctly. You may remove that changes.

MatteoLdx:
> @MatteoLdx I think that changing `SCRLfs` to `SC` is a little bit to early, because the CI is not ready yet to check that correctly. You may remove that changes.

Yes, I quickly realised it was a bad idea. Removed !

beesdoo_worker_status_shift_swap/models/task.py
MatteoLdx: Yes, setting the status to None prevents the counter update, so there's no need to change the counter manually. It is better than just setting data["sr"] to 0 because it also prevents the update of the next alert date, which was not the case before this commit.

beesdoo_shift/models/task.py
MatteoLdx: Done !

remytms (approved):
LGTM. We put in test and wait for comments.

remytms:
@MatteoLdx Can you just check if tests works on your local database ?

PR #423, opened 2022-06-29
[FIX] b_website_shift: move modals to avoid position problems

MatteoLdx:
## Description

Move website modals to fix inaccessibility problem on some mobile devices.


remytms (approved):
LGTM

PR #431, opened 2022-07-12
Change the beesdoo_print_label module

FlyearthR:
## Description
Change the template to be more readable.

## Checklist before approval

- [ ] Change log snippet is present. (?)


beesdoo_print_label/__manifest__.py
```diff
@@ -18,4 +18,6 @@
         "data/product_print_category.xml",
     ],
     "installable": True,
+    "application": True,
```
robinkeunen, line 21: An application covers a new functional scope and gets its entry point in the menu. This is not an application but only an addon.

robinkeunen (changes requested):
Looks good except for a small comment. I'll deploy this so that Virginie can test. Can you also check the precommit checks ?

vdewulf:
@robinkeunen 
Tested on bees-test
See screenshot.
The product name is larger but it's quickly truncated (the text goes on a second line and we see the top of the letters only). I don't know what we can do about it... It seems to have some space above the first line, maybe possible to get the name displayed in big letter on 2 full lines? This will need to be discussed with users.
The "fournisseur" text is removed (but the provider is still displayed and is larger) : ok
The "consigne" part is moved lower : ok
The "poids" and "prix/l ou prix/kg" is removed : ok
The price per unit: the "/[display_unit.name] is not added : NOT OK

[test_etiquettes.pdf](https://github.com/beescoop/Obeesdoo/files/9297921/test_etiquettes.pdf)

Could you correct the NOT OK line (about the name, let's keep it like this and discuss with users) before next intercoop meeting so that I can send it to the participants beforehand?
Thanks!

vdewulf:
@robinkeunen 
I discussed with Matthieu (bloum), we think that the product name should be on 2 lines (and "big letters"). If you have the courage to take this :)
Maybe with @remytms's help?

robinkeunen:
Closing in favor of https://github.com/beescoop/Obeesdoo/pull/445

carmenbianca:
@robinkeunen this PR is still enabled on test. #445 is not. please look at this when you're back.

PR #289, opened 2022-02-24
[ADD] member_card: add member card report

tfrancoi:
This module add the ability to print the member card
from the contact to be use with beesdoo_base

PR #432, opened 2022-07-12
Add customer_member_card module

FlyearthR:
## Description
Allows to define a specific logo for the member cards

## Checklist before approval

- [ ] (?)Change log snippet is present.


customer_member_card/__manifest__.py
```diff
@@ -0,0 +1,28 @@
...
+
+    'depends': ['base'],
+
+    'data': ['views/member_card_logo.xml'],
+
+    'license': 'AGPL-3',
+
+    'installable': True,
+
+    'application': True,
```
robinkeunen, line 25: As for the other PRs, this is not an application. Can you also remove the empty lines ?

customer_member_card/models/models.py
```diff
@@ -0,0 +1,7 @@
+#coding:utf-8
+
+from odoo import models, fields
+
+class CustomerMemberCard(models.Model):
+    _inherit = "res.company"
+    member_card_logo = fields.Char(string="Member Card Logo")
```
robinkeunen, line 7: You'll rather need to use a `fields.Binary` to store the logo.

robinkeunen: Alternatively, you could use the company logo, [check this](https://github.com/odoo/odoo/blob/67461b4ca1a9c2eb8ea1c0cc4bc39d91cf0d2823/odoo/addons/base/models/res_company.py#L39) .

customer_member_card/models/models.py
```diff
@@ -0,0 +1,7 @@
+#coding:utf-8
```
robinkeunen, line 1: This is not needed in python3

robinkeunen (changes requested):
The template for the card from [Thibault's PR](https://github.com/beescoop/Obeesdoo/pull/289/files) is missing. Other comments in the code. Good start though.

ping @tfrancoi

PR #435, opened 2022-07-18
[UPD] translation files

MatteoLdx:


carmenbianca (approved):
translations are kinda weird, but whatever.

PR #436, opened 2022-07-18
12.0 [FIX] access rights

MatteoLdx:
## Description

Fix an access right problem when trying to get a config parameter.

## Odoo task (if applicable)

T8921


carmenbianca (approved):
LGTM

PR #437, opened 2022-07-18
12.0 [IMP] beesdoo_shift_swap

MatteoLdx:
## Description

Improvements and bugfixes, mostly for exchange modules.

T9061

## Checklist before approval

### DIFF
--- a/beesdoo_shift/models/res_partner.py
+++ b/beesdoo_shift/models/res_partner.py
@@ -229,6 +229,10 @@ class ResPartner(models.Model):
             task_template = self.env["beesdoo.shift.template"].search(
                 [("worker_ids", "in", self.id)], limit=1
             )
+
+            if not task_template:
+                return generated_shifts, []
+
             main_shift = self.env["beesdoo.shift.shift"].search(
                 [
                     ("task_template_id", "=", task_template[0].id),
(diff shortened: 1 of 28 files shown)


beesdoo_shift_swap/models/task.py
```diff
@@ -34,3 +34,9 @@ def subscribe_shift_as_solidarity(self):
     def _compute_is_solidarity(self):
         for rec in self:
             rec.is_solidarity = bool(rec.solidarity_offer_ids)
+
+    def cancel_solidarity_offer(self):
+        self.ensure_one()
+        if self.is_solidarity:
+            return self.solidarity_offer_ids[0].cancel_solidarity_offer()
```
carmenbianca, line 41: Maybe I'm naive, but what about all the other solidarity offer ids in the recordset? i.e. `solidarity_offer_ids[1:]`

beesdoo_shift_swap/views/shift_swap_proposale.xml
```diff
@@ -15,7 +15,7 @@
                     class="oe_highlight"
                     type="object"
                     groups="beesdoo_shift.group_shift_management"
-                    attrs="{'invisible': [('working_mode','!=','regular'),('state','!=','suspended'),('state','!=','resigning')]}"
+                    attrs="{'invisible': ['|',('working_mode','!=','regular'),('state','in',['unsubscribed','resigning'])]}"
```
carmenbianca, line 18: good catch!

beesdoo_shift/models/res_partner.py
```diff
@@ -229,55 +229,58 @@ def get_next_shifts(self):
...
-            main_shift = self.env["beesdoo.shift.shift"].search(
-                [
-                    ("task_template_id", "=", task_template[0].id),
-                    ("start_time", "!=", False),
-                    ("end_time", "!=", False),
-                ],
-                order="start_time desc",
-                limit=1,
-            )
+            if task_template:
```
carmenbianca: Instead of indenting everything with `if task_template`, could we do something like:

```python
task_template = [...]
if not task_template:
    return generated_shifts, []
main_shift = [...]
```

Because: Flat is better than nested.

beesdoo_website_shift_swap/views/exchange_templates.xml
```diff
@@ -318,9 +318,15 @@
                     />
                 </form>
             </p>
+            <p t-if="same_timeslot" class="text-center">
+                <a
+                    class="btn btn-primary"
+                    href="/my/shift/possible/shift"
+                >No shifts suits me</a>
```
carmenbianca:
```suggestion
                >No shift suits me</a>
```

carmenbianca (changes requested):
Mostly LGTM, some comments.

beesdoo_shift_swap/models/task.py
MatteoLdx: There should never be more than one, this One2many is there to link the shift to the solidarity offer (the solidarity offer has a Many2one field containing the shift)

beesdoo_website_shift_swap/views/exchange_templates.xml
MatteoLdx: I agree with that, I just copied the other same sentences because I was too lazy to change them all

beesdoo_shift/models/res_partner.py
MatteoLdx: Done

beesdoo_shift_swap/models/task.py
carmenbianca: Ah, it's a fake One2one relationship. Consider https://odoo-development.readthedocs.io/en/latest/dev/py/one2one.html

polchampion:
@remytms Can you follow-up on this PR?

beesdoo_shift_swap/models/task.py
remytms: In all other code of obeesdoo, the one2one relation are done like Mattéo does. I found that the hack provided is as complicated as the solution here. So we may leave it like this.

beesdoo_website_shift_swap/views/exchange_templates.xml
remytms: Fixed everywhere. :)

MatteoLdx: Thanks !

beesdoo_shift_swap/models/task.py
carmenbianca: OK!

carmenbianca (approved):

PR #443, opened 2022-08-11
[FIX] beesdoo_shift_attendance: remove has_accounting entries

robinkeunen:
Breaks the update after module splits. The field is not
displayed nor used in the inheriting view


remytms (commented):
LGTM

PR #447, opened 2022-09-06
[12.0] fix cooperator type computation 

### LINKED ISSUE #446: [IMP] beesdoo_easy_my_coop : Cooperators without `birthdate` should be "Worker and Eater"
As is, in a subscription request, if the partner's `birthdate` is not defined, `eater` will be "Eater"
To be : if `birthdate` is False, let `eater` be "Worker and Eater" 
Else if `birthdate` is True, continue as is.

robinkeunen:
[FIX] beesdoo_easy_my_coop: remove cooperator_type overide

    Before refactoring of beesdoo_base, beesdoo_base and cooperator
    defined cooperator_type. beesdoo_base defined it with compute=None.
    beesdoo_easy_my_coop depended on cooperator and beesdoo base.
    cooperator was loaded later and thus the compute function was set.

    After refactoring, cooperator_type was moved from beesdoo_base to
    beesdoo_easy_my_coop and thus overrode cooperator definition.
    compute=None was set to the fields.

    With this fix, we restore the use of the compute function and the
    get_share_type from cooperator module.

[FIX] beesdoo_easy_my_coop: check age is set

    When no birthdate is set on partner, age is 0.
    In that case, partner is not minor and should be eater_worker.


beesdoo_easy_my_coop/models/coop.py
```diff
@@ -10,7 +10,7 @@ def get_eater_vals(self, partner, share_product_id):
         vals = {}
         eater = share_product_id.eater
 
-        if partner.is_company or partner.age < self._majority:
+        if partner.is_company or 0 < partner.age < self._majority:
```
huguesdk: what about a child that is not yet 1 year old? i suggest to check whether `partner.birthdate_date` is truthy first.

```suggestion
        if partner.is_company or (partner.birthdate_date and partner.age < self._majority):
```

also, since you’re fixing things here, `_majority` is a “constant”, so it should be written in capitals. personally, i would write it outside of the class, since it is not a property of the `subscription.request` model.

huguesdk (changes requested):

beesdoo_easy_my_coop/models/coop.py
huguesdk: [age compute function](https://github.com/OCA/partner-contact/blob/12.0/partner_contact_birthdate/models/res_partner.py#L20), for reference.

robinkeunen: I do like your solution better 👍

robinkeunen (commented):
@huguesdk Can you check the refactored code ?

beesdoo_easy_my_coop/models/subscription_request.py
```diff
@@ -22,3 +24,23 @@ def get_required_field(self):
...
+        if partner.is_company or (partner.birthdate_date and partner.age < _MAJORITY):
+            eater = "eater"
+
+        return {"eater": eater}
+
+    @api.multi
+    def validate_subscription_request(self):
+        self.ensure_one()
+
+        invoice = super(SubscriptionRequest, self).validate_subscription_request()[0]
```
huguesdk:
```suggestion
        invoice = super().validate_subscription_request()[0]
```
this is python 3.

beesdoo_easy_my_coop/models/subscription_request.py
```diff
@@ -22,3 +24,23 @@ def get_required_field(self):
         if company.info_session_confirmation_required:
             required_fields.append("info_session_confirmed")
         return required_fields
+
+    def get_eater_vals(self, partner, share_product_id):
+        eater = share_product_id.eater
+
+        if partner.is_company or (partner.birthdate_date and partner.age < _MAJORITY):
```
huguesdk, line 33: a “why” comment would be useful.

huguesdk (approved):
woohoo, tests and all! just these 2 comments and we’re good. 👍

beesdoo_easy_my_coop/models/coop.py
carmenbianca: Agree with @huguesdk

beesdoo_easy_my_coop/models/subscription_request.py
```diff
@@ -1,7 +1,9 @@
 # Copyright 2019 Coop IT Easy SC
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
+
+_MAJORITY = 18
```
carmenbianca, line 6: Should this be a configurable parameter?

carmenbianca (changes requested):
Also: Commit message is wrong.

```
 [FIX] beesdoo_easy_my_coop: check age is set

When no birthdate is set on partner, age is 0.
In that case, partner is not minor and should be eater_worker.
```

Change to 'assume that partner is not a minor if birth date is not defined' or some such.

beesdoo_easy_my_coop/models/subscription_request.py
robinkeunen: future is now

beesdoo_easy_my_coop/models/subscription_request.py
robinkeunen: It could be but I won't add now 😇

robinkeunen:
Fixed the commit message @carmenbianca

carmenbianca (approved):
LGTM

PR #411, opened 2022-06-09
12.0 split beesdoo product product main supplier

victor-champonnois:
fix https://github.com/beescoop/Obeesdoo/issues/392
fix https://github.com/beescoop/Obeesdoo/issues/409
fix https://github.com/beescoop/Obeesdoo/issues/391

Split part of beesdoo_product into:
- product_main_supplier
- sale_suggested_price
- sale_edit_price_wizard (renamed to sale_adapt_price_wizard)

The dependency graph is the following :
`sale_edit_price_wizard` -> `sale_suggested_price` -> `product_main_supplier`

Todo : merge https://github.com/beescoop/Obeesdoo/pull/320 

Migration is done in modules with [this method](https://github.com/coopiteasy/vertical-cooperative/issues/233#issuecomment-1141304743)

PR #459, opened 2022-10-19
[FIX] beesdoo_product: main_seller_id search

remytms:
## Description

The Main Seller can't be sorted in tree view and also the value of the column `main_seller_id` in the database is not updated anymore, causing issue to external tools that rely on this column.

## Odoo task (if applicable)

[task](https://gestion.coopiteasy.be/web#id=9305&model=project.task&view_type=form&menu_id=)

## Checklist before approval

### DIFF
--- a/beesdoo_product/models/beesdoo_product.py
+++ b/beesdoo_product/models/beesdoo_product.py
@@ -1,9 +1,9 @@
 # Copyright 2020 Coop IT Easy SCRL fs
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
+import datetime
 import logging
 import uuid
-from datetime import date
 
 from odoo import api, fields, models
 from odoo.exceptions import UserError, ValidationError
@@ -64,19 +64,21 @@ class BeesdooProduct(models.Model):
         translate=True,
     )
 
-    top_supplierinfo_id = fields.Many2one(
+    main_supplierinfo_id = fields.Many2one(
         comodel_name="product.supplierinfo",
-        compute="_compute_main_seller_id",
+        compute="_compute_main_supplierinfo_id",
         store=True,
     )
     main_seller_id = fields.Many2one(
         string="Main Seller",
         comodel_name="res.partner",
-        related="top_supplierinfo_id.name",
+        related="main_supplierinfo_id.name",
+        store=True,
     )
     main_seller_id_product_code = fields.Char(
         string="Main Seller Product Code",
-        related="top_supplierinfo_id.product_code",
+        related="main_supplierinfo_id.product_code",
+        store=True,
     )
 
     display_unit = fields.Many2one("uom.uom")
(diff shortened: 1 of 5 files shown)


beesdoo_product/models/beesdoo_product.py
```diff
@@ -64,19 +64,21 @@ class BeesdooProduct(models.Model):
         translate=True,
     )
 
-    top_supplierinfo_id = fields.Many2one(
+    main_supplierinfo_id = fields.Many2one(
         comodel_name="product.supplierinfo",
-        compute="_compute_main_seller_id",
+        compute="_compute_main_supplierinfo_id",
         store=True,
     )
```
remytms, line 71: I name it wrong first time. As it is not the first supplierinfo but realy the main one as it's computed via the `_get_main_supplier_info()` method.

beesdoo_product/models/beesdoo_product.py
```diff
@@ -163,17 +165,22 @@ def _compute_scale_sale_uom(self):
...
+        """
 
-        # supplierinfo w/o date_start come first
-        def sort_date_first(seller):
+        def sort_date_asc(seller):
             if seller.date_start:
                 return seller.date_start
             else:
-                return date.max
+                return datetime.date.max
```
remytms, line 181: Changed this because there is fields or variable that can be named `date` and so I didn't get it directly that it was just the method from the `datetime.date` package. Like this it is clearer, IMHO.

beesdoo_product/models/beesdoo_product.py
```diff
@@ -219,15 +226,9 @@ def generate_barcode(self):
...
+    def _compute_main_supplierinfo_id(self):
         for product in self:
-            # todo english code Calcule le vendeur associé qui a la date de
-            #  début la plus récente et plus petite qu’aujourd’hui fixme
-            #   could product.main_seller_id be used instead? it seems that
-            #   “seller” and “supplier” are used interchangeably in this
-            #   class. is this on purpose?
-            sellers_ids = product._get_main_supplier_info()
-            product.top_supplierinfo_id = sellers_ids and sellers_ids[0] or False
+            product.main_supplierinfo_id = product._get_main_supplier_info()
```
remytms, line 231: Comments are now in the `_get_main_supplier_info()` method.

remytms: No need to double check that there is only one supplierinfo as it is part of the `_get_supplierinfo()` method.

polchampion:
@victor-champonnois @remytms Ok pour vous deux de décider à 2 cette aprèm qu'est-ce qu'on rebase sur quoi ? Pingez-moi si besoin d'input. L'idée est de déployer le tout le 31, mais j'ai demandé à Bart si c'était ok d'attendre encore 10 jours. Je vous laisse voir ce qui fait techniquement le plus de sens à ce stade.

victor-champonnois:
All changes incorporated in this PR https://github.com/beescoop/Obeesdoo/pull/411

PR #461, opened 2022-10-27
[ADD] cooperator_worker_force

robinkeunen:
On partner form, tick "Force Worker" box to set this partner
as a worker even if they are not an effective cooperator.

## Checklist before approval


robinkeunen:
@tfrancoi I moved all the worker_store relayed stuff in a separated module. This should not affect polln but it might be a good idea to have a look.

robinkeunen:
@polchampion This is deployed on test server (except demain). The _Force Worker_ functionality was moved to a new module `cooperator_worker_force`. I suggest to test 

1) that the computation of is_worker is still correct before installing the new module
2) that the computation of is_worker takes `worker_store` (_Force Worker_) into account after installing the module.

polchampion:
@robinkeunen Sucessfully tested on lepedalo-test 

- [x]  the computation of is_worker is correct before and after installing the new module
- [x] the computation of is_worker takes `worker_store` (_Force Worker_) into account after installing the module.
- [x] the module seems compatible with share operations involving change in share that allow owner to work and not.

cooperator_worker_force/models/res_partner.py
```diff
@@ -0,0 +1,36 @@
+# Copyright 2022 Coop IT Easy SC
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
+
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+
+    _inherit = "res.partner"
+    worker_store = fields.Boolean(
```
victor-champonnois, line 10: Shouldn't this field be name `worker_force` ?

victor-champonnois (commented):
Just one comment, I am not finished ;)

remytms (approved):

victor-champonnois (approved):
LGTM.

About the whole worker/allow_working logic, wouldn't it be simpler to set `is_worker` as a related field from the share's `allow_working` ? 
- The field related field could be stored, so we wouldn't have to write search methods
- The field could be not readonly, so it could be modified on the interface.

robinkeunen:
I would be a simpler design indeed but a cooperator can have multiple shares so a simple related field won't do. We could have a "main share" field like "main seller" but it would be part of bigger refactoring which is already part of issue https://github.com/OCA/cooperative/issues/18

cooperator_worker_force/models/res_partner.py
robinkeunen: Okay, will do. I was feeling lazy ^^

victor-champonnois:
> I would be a simpler design indeed but a cooperator can have multiple shares so a simple related field won't do. We could have a "main share" field like "main seller" but it would be part of bigger refactoring which is already part of issue [OCA/cooperative#18](https://github.com/OCA/cooperative/issues/18)

Ah yes, forgot about that :+1:

cooperator_worker_force/models/res_partner.py
robinkeunen: Actually, I'll leave it for [later](https://github.com/beescoop/Obeesdoo/issues/471): I'm not sure how migration scripts will behave since this module will be installed at production deployment.

PR #464, opened 2022-10-28
[REF] beesdoo_stock: split module

robinkeunen:
https://github.com/OCA/stock-logistics-workflow/pull/1119


beesdoo_stock/models/stock.py
```diff
@@ -1,28 +0,0 @@
...
-                "view_mode": "list",
-                "res_model": "product.template",
-                "view_id": False,
-                "target": "current",
-                "type": "ir.actions.act_window",
-                "context": ctx,
-                "nodestroy": True,
-                "res_id": ctx["articles"],
-                "domain": [("id", "in", ctx["articles"])],
-            }
```
victor-champonnois, line 28: I don't understand, this is not the same code as in https://github.com/OCA/stock-logistics-workflow/pull/1119/files#diff-df167430c1f8baf5c14a15db17f7fc03019412c41df18fee8042b289c605ffc2. Is it just a split or did you change some stuff ?

robinkeunen: I recoded it the proper way indeed.

victor-champonnois (approved):
LGTM

PR #468, opened 2022-10-31
12.0 rename beesdoo pos reporting

### LINKED ISSUE #346: Rename `beesdoo_pos_reporting` to `pos_order_count_store`
According to https://github.com/coopiteasy/vertical-cooperative/issues/204, 'add api.dependencies on compute function'.

victor-champonnois:
## Description
fix https://github.com/beescoop/Obeesdoo/issues/346

## related PR

- https://github.com/OCA/pos/pull/883

## Checklist before approval


robinkeunen (approved):
Looks good but could it be moved to the oca directly?

victor-champonnois:
@robinkeunen according to https://user-images.githubusercontent.com/17589077/149927982-58f1b329-9d15-48d3-b331-77741f7c2e87.png it should be moved to cie-pos, plans have changed ?

robinkeunen:
My point of view has changed now yes. Maybe @legalsylvain could tell us whether he thinks `pos_order_count_store` would be welcome on oca/pos ?

The description.rst : 

> Store pos_order_count to improve reporting.

pos_order_count_store/models/res_partner.py
```diff
@@ -0,0 +1,12 @@
+from odoo import api, fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    pos_order_count = fields.Integer(store=True)
```
legalsylvain: I fear that if we make an update all, the update of point_of_sale module drop the column, and then, the update of the pos_order_count_store module recompute it. Am I wrong ?

legalsylvain: (I remember that I developed one time a module store = True -> store =False, and that was problematic.

legalsylvain (commented):
Hi. 

Thanks for the ping. I think that the module ``pos_order_count_store`` has its place under OCA umbrella. (generic and add features), as far as the "update" question is OK.

I think that if yes, you should add a pre-init-hook script to compute by SQL the stored field.
Otherwise, installing this module on existing databases (the GRAP one, for instance) will take hours for sure.

pos_order_count_store/models/res_partner.py
victor-champonnois: Thank you for your reply

> I fear that if we make an update all, the update of point_of_sale module drop the column, and then, the update of the pos_order_count_store module recompute it. Am I wrong ?

It's not what I observe, when updating point_of_sale (and then pos_order_count_store), the field was never recomputed.

Noted for the SQL Query in pre-init hook, I'll do that !

victor-champonnois:
- I added the pre_init_hook
- @polchampion OK to test on coopeco-test and lesptitpots-test. To be tested : "Store pos_order_count to improve reporting." (check that the number of pos_order per partner is OK)

robinkeunen (changes requested):
Thx for the pre-init hook 👍 Can you rebase, squash and move the module to oca/pos ?

victor-champonnois:
@robinkeunen LGTM, should be merged for next MEP, so that pos_counter_store is installed

victor-champonnois:
@robinkeunen can you approve the PR ?

robinkeunen (approved):

polchampion:
@victor-champonnois Sucessfully tested on coopeco-test : number of pos orders for one partner seems right (smart button on profile + pos order dynamic table view + filter on all orders for that partner.

PR #469, opened 2022-11-01
[REF] split beesdoo_product_usability->product_main_supplier

### LINKED ISSUE #360: Merge `beesdoo_product_usability` into `product_main_supplierinfo`
Also backport `product_main_supplierinfo` to 12.0.

victor-champonnois:
## Description
fix https://github.com/beescoop/Obeesdoo/issues/360

See comments in the code 

## Checklist before approval


product_main_supplier/views/product_template_views.xml
```diff
@@ -12,4 +12,48 @@
         </field>
     </record>
 
+    <!-- Should this view inherit and modify rather than replace? -->
```
robinkeunen: All these views were meant to simplify the view definitions which became a bit messy.

The fields `main_seller_id_price` and `main_seller_id_minimum_qty` should come in this module, the big view replacement can stay in `beesdoo_product_usability` for the time beeing.

product_main_supplier/views/product_template_views.xml
```diff
@@ -12,4 +12,48 @@
...
+                <field name="stock_coverage" />
+                <field name="daily_sales" />
+                <field name="range_sales" />
+                <field name="list_price" />
+                <field name="effective_sale_price" />
+            </tree>
+        </field>
+    </record>
+
+    <!-- These look like customizations rather than functionalities,
```
robinkeunen: This can stay in `beesdoo_product_usability`

robinkeunen (changes requested):

victor-champonnois:
@robinkeunen thanks for the reply

Also check my comments (copied from the issue):

>Oups, my PR merge beesdoo_product_usability into obeesdoo's product_main_supplier. I misundertood.
>I have the impression that [product_main_supplierinfo](https://github.com/OCA/product-attribute/blob/14.0/product_main_supplierinfo/models/product_product.py) is doing more or less the same thing as Obeesdoo's product_main_supplier.

>Should we merge product_main_supplier into product_main_supplierinfo ?
If not, if we merge beesdoo_product_usability into product_main_supplierinfo, would not be able to use it because we would have to install product_main_supplierinfo, while already having product_main_supplier.

I think we should go on and think later about whether we should merge `product_main_supplier` with `product_main_supplierinfo` and how to do it. What do you think ?

robinkeunen:
@victor-champonnois I agree with you, let's move on and keep `product_main_supplier` in v12. Can you add note pointing to [product_main_supplierinfo](https://github.com/OCA/product-attribute/blob/14.0/product_main_supplierinfo/models/product_product.py) in `product_main_supplier`'s roadmap.rst ?

product_main_supplier/views/product_template_views.xml
victor-champonnois: done

product_main_supplier/views/product_template_views.xml
victor-champonnois: done

victor-champonnois:
@robinkeunen ready to review again

robinkeunen (approved):

PR #472, opened 2022-11-02
[FIX] update test-requirements

robinkeunen:
These module have been moved to the oca/cooperative repository.


carmenbianca (approved):
oca_dependencies.txt can be removed. otherwise lgtm

PR #498, opened 2023-01-31
[FIX] barcode rule priority

victor-champonnois:
beesdoo_barcode_rule which was high on the priorities, it has been replaced by a new rule with sequence=45. This resulted in error because another rule with higher priority was used instead. Therefore we increase the priority of this rule.

This can only be tested on new DB because the record has noupdate.
## Checklist before approval


carmenbianca (approved):
LGTM

PR #506, opened 2023-03-27
[16.0] [MIG] shift_portal

victor-champonnois:
## Odoo task (if applicable)

https://gestion.coopiteasy.be/web?debug=true#id=10308&action=475&active_id=492&model=project.task&view_type=form&menu_id=536

## Checklist before approval


shift_portal/views/res_config_views.xml
```diff
@@ -7,25 +7,128 @@
     <record id="view_website_shift_config_irregular" model="ir.ui.view">
         <field name="name">Website Shift Settings Irregular Worker</field>
         <field name="model">res.config.settings</field>
-        <field name="inherit_id" ref="website.res_config_settings_view_form" />
+        <field name="inherit_id" ref="shift.res_config_settings_shift_view_form" />
+        <!-- priority 21 to put it after website (priority 20) -->
+        <field name="priority" eval="21 " />
```
remytms: You may want to chose a higher priority so that if something should come between the two, it's possible. Something like 30.

But this change may have undesired effect if there is some other view with other priority, so let's not change it now.

remytms (approved):
LGTM.

shift_portal/views/res_config_views.xml
victor-champonnois: OK

carmenbianca: done

PR #511, opened 2023-05-09
16.0 [ADD] shift_beneficiary

victor-champonnois:
## Description
Add a beneficiary field on shifts (see description.rst)

## Odoo task (if applicable)
https://gestion.coopiteasy.be/web#id=10311&action=475&active_id=455&model=project.task&view_type=form&menu_id=

## Checklist before approval


shift_beneficiary/controllers/main.py
```diff
@@ -0,0 +1,42 @@
...
+        if selected_beneficiary:
+            res = res and task_template.beneficiary == selected_beneficiary
+        return res
+
+    def get_selected_beneficiary(self):
+        beneficiary_model = request.env["res.partner"]
+        if "beneficiary" in request.session:  # a filter has been applied
+            beneficiary_id = request.session["beneficiary"]
+            if beneficiary_id:  # the filter is not empty
+                return beneficiary_model.browse(int(beneficiary_id))
```
remytms: If using `int(beneficiary_id)`, you should be absolutely sure that it cannot throw an error, else the website will just render an error 500.

shift_beneficiary/controllers/main.py
```diff
@@ -0,0 +1,42 @@
...
+        return beneficiary_model
```
remytms: There is indentation with successive if which may not be easy to read. So, it's ok to leave it like this, but I may suggest the following:

```python
try:
    beneficiary_id = int(request.session("beneficiary", 0))
except ValueError:
    beneficiary_id = 0
return request.env["res.partner"].browse(beneficiary_id).exists()
```

shift_beneficiary/controllers/main.py
```diff
@@ -0,0 +1,42 @@
...
+            request.env["res.partner"].sudo().search([("is_beneficiary", "=", True)])
+        )
+        res.update({"beneficiary_list": beneficiary_list})
+        res.update({"selected_beneficiary": self.get_selected_beneficiary()})
+        return res
+
+    @http.route("/shift_irregular_worker", auth="public", website=True)
+    def public_shift_irregular_worker(self, **kw):
+        if request.httprequest.method == "POST":
+            request.session["beneficiary"] = kw["beneficiary"]
```
remytms: Considering my previous note. Here you cannot be sure that the content of `kw["beneficiary"]` is always a number. POST values are user input, so never trust it.

shift_beneficiary/models/shift_shift.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class ShiftShift(models.Model):
+    _inherit = "shift.shift"
+
+    template_beneficiary = fields.Many2one(
+        "res.partner",
+        related="task_template_id.beneficiary",
+        store=True,
+    )
```
remytms, line 11: With this design, you cannot create a shift linked to a beneficiary without having a template linked to this shift.

In the current design, you can create a shift that is not linked to a task_template.

Having `beneficiary` linked only to the task_template_id, you cannot create a shift with a beneficiary that is not linked to an existing `task_template_id`.

I think for example at the `task_type_id` field, which exists in the template and also on the shift. Task Type is copied from the template at the generation of the shift.

Also, if in the future a template is modified, all the already generated shift will see their beneficiary to change, even if in the past it was not the case.

remytms (changes requested):
Warning about the way beneficiary is implemented on the `shift.shift` model and about the trust of the user input in the POST section. Except that no changes requested.

shift_beneficiary/controllers/main.py
victor-champonnois: OK, so this issue would be taken care of by the change you proposed in the `get_selected_beneficiary` method ?

shift_beneficiary/models/shift_shift.py
victor-champonnois: I see, do you think I should rewrite the code so that beneficiary_id behaves like task_type_id ? Or can we go with this design and refactor afterward if needed ?

shift_beneficiary/controllers/main.py
victor-champonnois: Thanks for the proposal, I applied it. I just added `get` after `session` that you forgot I think. 
What does the exists() method do ?

remytms: Indeed, the `get` method stay in my keyboard. :p

The [exists method](https://www.odoo.com/documentation/12.0/developer/reference/orm.html?highlight=exists#odoo.models.Model.exists). Basicly it returns only the existing records from the recordset.

shift_beneficiary/controllers/main.py
remytms: Yes, that's a way of doing it. :)

shift_beneficiary/models/shift_shift.py
remytms: This should be discussed with Pol. Because it's a matter of edge case and usage of the module. I can be part of the discussion to explain.

But the refacto can be done later, it will just need a migration script. You may write a note in the README about this limitation for now.

remytms (approved):
Just squash your commit or rename your fixup commit to track the changes made. And think about adding a note in the readme about limitation if you have time. :)

victor-champonnois:
@remytms done ! The base branch (mig-shift_portal) is not merged, so I don't think we should merge this branch yet

shift_beneficiary/models/shift_shift.py
polchampion: @remytms @victor-champonnois I think at some point we need to rethink properly the relationship shift.template > shift.shift, notably : 
- what's the status on creating shifts not linked to a template (originally allowed, except it caused errors in the portal display of future shifts)
- what infos shifts inherit from template and which should be duplicated or linked to the template
- which features update shift info after template modification, and do they make sense (see discussions between Pol and Yo about this)

I'm copying these notes to the shift analysis task (to be done in 6+ months by Rémy).

PR #516, opened 2023-06-01
[12.0][FIX] eater: Add restraints on parent and child eaters

carmenbianca:
## Odoo task (if applicable)

https://gestion.coopiteasy.be/web#id=10654&model=project.task&view_type=form&menu_id=536

## Checklist before approval


eater/models/partner.py
```diff
@@ -18,9 +18,53 @@ class Partner(models.Model):
         domain=[("customer", "=", True), ("eater", "=", "eater")],
     )
     parent_eater_id = fields.Many2one(
-        "res.partner", string="Parent Worker", readonly=True
+        "res.partner",
+        string="Parent Worker",
+        readonly=True,
+        domain=[("customer", "=", True), ("eater", "=", "worker_eater")],
```
victor-champonnois: I think the domain condition on customer should be removed. There is some examples in our databases that the workers/eaters and the eaters are not customer (see on demain.test). So I think it will cause some worker/eater to not appear in the dropdown list.

eater/models/partner.py
```diff
@@ -18,9 +18,53 @@ class Partner(models.Model):
...
+            ) and not partner.customer:
+                raise ValidationError(
+                    _("%s must be a customer to be an eater or a parent worker.")
+                    % partner.name
+                )
+            if partner.parent_eater_id and not partner.parent_eater_id.customer:
+                raise ValidationError(
+                    _("%s must be a customer to be a parent worker.")
+                    % partner.parent_eater_id.name
+                )
```
victor-champonnois: again here I fear that it will clash with the state of the databases. I would say we drop this constraint. 
If we want to implement this, we should write a migration script to convert all eaters and worker/eaters as customer, and also ensure that the current process that creates the partners always sets them as customer. This is I think beyond the scope of this bug fix.

eater/views/partner.xml
```diff
@@ -7,11 +7,10 @@
...
         <field name="arch" type="xml">
 
+
+
             <xpath expr="//notebook" position="inside">
-                <page
-                    string="Eaters"
-                    attrs="{'invisible': ['|', ('customer', '=', False), ('eater', 'not in', ('worker_eater'))]}"
-                >
+                <page string="Eaters" attrs="{'invisible': [('customer', '=', False)]}">
```
victor-champonnois: again, see comment above

victor-champonnois (changes requested):
Thanks for the PR, some issue with adding a is_customer constraint. Could not do a functional test since the situation is very hard to reproduce.

carmenbianca:
@victor-champonnois I started removing the relevant code, but then I began noticing a lot of other code that I had **not** touched that also acted on the assumption that `customer=True` for parent/child eaters.

Not wanting to delete code unrelated to this PR, I just wrote the migration script to fix this.

eater/__manifest__.py
```diff
@@ -11,7 +11,7 @@
     "author": "BEES coop - Cellule IT, Coop IT Easy SC",
     "website": "https://github.com/beescoop/Obeesdoo",
     "category": "Sales",
-    "version": "12.0.1.0.0",
+    "version": "12.0.3.0.0",
```
carmenbianca: I think I incorrectly bumped this two versions.

eater/models/partner.py
carmenbianca: Fixed by e218bf6dcc31e4cb34bf7d51b76d8c4dd33f2d98

eater/models/partner.py
carmenbianca: Migration script written in e218bf6dcc31e4cb34bf7d51b76d8c4dd33f2d98

> and also ensure that the current process that creates the partners always sets them as customer

Already the case in `_new_eater`

carmenbianca:
Odoo 13.0 removes the customer field. In #521 we're getting rid of all customer field-related code. To facilitate a smoother migration (because we'll need to port these changes to the migration branch), I removed the added restrictions related to the customer field (+ I removed the pre-existing ('customer', '=', True) domain from a field).

I didn't touch any other customer field-related code.

victor-champonnois (approved):
LGTM, thanks for the changes :)

PR #518, opened 2023-06-20
[12.0][FIX] fix penalty setting resetting on update

huguesdk:
## description

fix `shift_worker_status.irregular_penalty` config parameter resetting to true when updating the module.

## odoo task

[t10727](https://gestion.coopiteasy.be/web#id=10727&model=project.task)


remytms (approved):
Thanks for the fix and the comment.

PR #521, opened 2023-08-03
[16.0] [MIG] eater

victor-champonnois:
internal task : https://gestion.coopiteasy.be/web#id=10884&action=475&active_id=492&model=project.task&view_type=form&menu_id=

I need technical input on two questions  : bca51764f593b3b9acd9bbd03e3b4e63da810fbb and 9902d8890c607fc0219cfe9494528566f8fec883

@polchampion Since the field "customer" has been removed in v13, I removed 
- the restriction on customer in the domain for adding new eaters.
- The invisibility of the field for non-customer contact : so the field Eater/worker is now visible on all contacts

This is not ideal, but I don't see a better solution. We could use the customer_rank field, which starts at zero and increases with the contact's orders. But this would mean that new contacts or contact that never shopped could not been added as Eater.  
Maybe we could make the field invisible for companies ?


polchampion:
@victor-champonnois I think it's fine as you did, let's keep it simple. Can you ping me when it's live in test?

cathLemb:
"Maybe we could make the field invisible for companies ?"
Not possible as bees for example uses eater field on companies.

carmenbianca:
I just realised that #516 is not is not included in this PR, and that the changes to customer will conflict with some changes. I'll fix up #516 and then rebase this work on top of that later down the line.

eater/static/font/Roboto-Bold.ttf
carmenbianca, line 1: This font (and some other static resources) are deleted in the 13.0 migration commit. Why?

eater/views/partner.xml
```diff
@@ -10,7 +10,7 @@
             <xpath expr="//notebook" position="inside">
                 <page
                     string="Eaters"
-                    attrs="{'invisible': ['|', ('customer', '=', False), ('eater', 'not in', ('worker_eater'))]}"
+                    attrs="{'invisible': [('eater', 'not in', ('worker_eater'))]}"
```
carmenbianca: This should be in the 13.0 migration commit.

eater/views/partner.xml
```diff
@@ -37,7 +37,7 @@
             </xpath>
 
             <field name="website" position="after">
-                <field name="eater" attrs="{'invisible': [('customer', '=', False)]}" />
+                <field name="eater" />
```
carmenbianca: idem

eater/models/partner.py
```diff
@@ -26,7 +26,7 @@ def write(self, values):
             if (
                 values.get("parent_eater_id")
                 and partner.parent_eater_id
-                and partner.parent_eater_id.id != values.get("parent_eater_id")
+                and partner.parent_eater_id != values.get("parent_eater_id")
```
carmenbianca: This makes no sense to me.

eater/models/partner.py
```diff
@@ -22,6 +22,8 @@ class Partner(models.Model):
     )
 
     def write(self, values):
+        # fixme: shouldn't the check on "values.get("parent_eater_id")" be before
+        # the for loop to avoid looping through all the contacts at every write?
```
carmenbianca: Sure.

eater/models/partner.py
```diff
@@ -37,8 +37,10 @@ def write(self, values):
...
                 )
-        # replace many2many command when writing on child_eater_ids to just
-        # remove the link
+        # fixme : is this still necessary ? In a many2many widget, removing an item
+        # is a "3" (unlink) command. In which case would this be a delete command ?
+        # replace many2many command when writing on child_eater_ids to unlink
+        # rather than delete
         if "child_eater_ids" in values:
             for command in values["child_eater_ids"]:
                 if command[0] == 2:
```
carmenbianca: I guess it depends on why this exists. The apparent goal to me is 'don't accidentally delete the linked child eaters when severing the link'. But why would the '2' command ever be sent when the '3' command is intended?

For context:

- `(2, id, _)` removes the link to and deletes the id related record.
- `(3, id, _)` removes the link to, but does not delete, the id related record.
  This is usually what you will use to delete related records on many-to-many
  fields.

---

In any case, if we keep this line, a test should be added.

carmenbianca (changes requested):
some notes

eater/static/font/Roboto-Bold.ttf
victor-champonnois: I could have removed it in a separate commit. It's just that it's not used and it makes no sense to keep this file here.

eater/views/partner.xml
victor-champonnois: done

eater/views/partner.xml
victor-champonnois: done

eater/models/partner.py
victor-champonnois: I would be in favor of removing it then. I don't see why this would be needed for this specific many2many field. Do you agree ?

carmenbianca: It's not a Many2many field. Not sure why the comment calls it a 'many2many command'.

I'm happy to drop this code also. Let's just make a note of it somewhere that the functional tester can be aware of this.

eater/models/partner.py
victor-champonnois: done

eater/models/partner.py
victor-champonnois: done

victor-champonnois: (note in the task)

robinkeunen (approved):

robinkeunen:
The migration commits should be stashed together.

huguesdk:
isn’t the module name too generic for such a specific module?

eater/models/partner.py
```diff
@@ -0,0 +1,81 @@
...
+                            "You are trying to assign %(eater)s as an eater to"
+                            " a new partner, but this eater is already"
+                            " assigned to %(old_partner)s. Please remove the link"
+                            " before creating a new one."
+                        )
+                        % {
+                            "eater": partner.name,
+                            "old_partner": partner.parent_eater_id.name,
+                        }
+                    )
```
carmenbianca: What is the use case here?

carmenbianca (changes requested):
I rebased this on top of the latest 12.0 to also include #516. I also applied a fix at the end, but I'm not entirely sure why the code that I fixed exists.

I'd like a confirmation that the tagged code is _useful_ before we merge this.

eater/models/partner.py
robinkeunen: You can't be the eater of two workers.

carmenbianca: But this would change the parent eater ("worker"), no? So after the operation, you still only have one parent.

robinkeunen: Indeed, that's weird.

eater/tests/test_eater.py
```diff
@@ -6,15 +6,15 @@ class TestEater(TransactionCase):
     def test_assign_parent_eater(self):
         self.assertTrue(
             self.env.ref("eater.eater2").write(
-                {"parent_eater_id": self.env.ref("eater.eater1")}
+                {"parent_eater_id": self.env.ref("eater.worker1").id}
```
robinkeunen: it's better practice to create records in the tests and leave demo data for demo.

eater/demo/eaters.xml
```diff
@@ -35,4 +35,20 @@
...
         <field name="is_company" eval="False" />
     </record>
+
+    <record id="worker1" model="res.partner">
+        <field name="name">Gijs Eaterdam</field>
+        <field name="firstname">Gijs</field>
+        <field name="lastname">Eaterdam</field>
+        <field name="eater">worker_eater</field>
+        <field name="is_company" eval="False" />
+    </record>
```
robinkeunen: It's better to overload existing partners than creating new ones.

robinkeunen (changes requested):

eater/models/partner.py
robinkeunen: Apparently, 8 years ago, this field was a many2many => removing the code.

robinkeunen: It dates back from 8 years ago. Let's trust ourselves and state that this is useless.  https://github.com/beescoop/Obeesdoo/commit/3f4d9f69c77c0bd9927e33b34182effb56d46568

carmenbianca (approved):
LGTM @robinkeunen

robinkeunen (approved):

PR #522, opened 2023-08-03
[16.0] [MIG] member_card

victor-champonnois:
task : https://gestion.coopiteasy.be/web#id=10992&action=475&active_id=492&model=project.task&view_type=form&menu_id=

I changed the behavior when creating a new card in 6da9ef27c5c286ffae8e0d2187351495e3af527d
- before, creating a new card would trigger the partner's `_compute_bar_code`  method, which would set the card's barcode to the partner. 
-  now the compute method is not triggered when creating a new card for unknown reasons. I then changed the behavior: 
    - the barcode is assigned to the partner on the creation of the member card
    - the fields on the member card are readonly, to force the creation of a new member card (and avoid barcode not being modified on the partner when it is modified on the member card).
@polchampion FYI (because it's a functional change) -> **Pol approved**

We looked at it with @robinkeunen  and we didn't found a better solution, do @carmenbianca or @huguesdk or @remytms have an opinion ?

Note : set to draft because we need to know how to deal with eater member cards  (komunigi decision)


member_card/demo/cooperators.xml
```diff
@@ -6,7 +6,6 @@
 <odoo>
     <record id="res_partner_cooperator_1_demo" model="res.partner">
         <field name="name">Fernand Peso</field>
-        <field name="customer" eval="True" />
```
carmenbianca: Why? Was this field removed?

member_card/report/member_card_template.xml
```diff
@@ -4,10 +4,11 @@
     <record id="report_member_card" model="ir.actions.report">
         <field name="name">Member Card</field>
         <field name="model">res.partner</field>
+        <field name="binding_model_id" ref="model_res_partner" />
         <field name="report_name">member_card.member_card_template</field>
         <field name="report_type">qweb-pdf</field>
         <field name="report_file">member_card.member_card_template</field>
-        <field name="print_report_name">Member Card</field>
+        <field name="print_report_name">'Member Card'</field>
```
carmenbianca: This strikes me as odd.

member_card/models/partner.py
```diff
@@ -39,4 +27,5 @@ def _new_card(self, reason, user_id, barcode=False):
         }
         if barcode:
             card_data["barcode"] = barcode
-        self.env["member.card"].create(card_data)
+        card = self.env["member.card"].create(card_data)
+        self.barcode = card.barcode
```
carmenbianca: I am not certain I like this solution.

1. If we're going to write the barcode string on the partner, we should do it in `member.card`'s create method, not here.
2. I think we should research a way to make the compute method work.

In any case, making `member.card` readonly seems like a fairly sound idea.

member_card/views/member_card.xml
```diff
@@ -0,0 +1,20 @@
...
+                    <field name="valid" readonly="1" />
+                    <field name="barcode" readonly="1" />
+                    <field name="partner_id" readonly="1" />
+                    <field name="responsible_id" readonly="1" />
+                    <field name="end_date" />
+                    <field name="comment" />
+                </group>
+            </form>
+        </field>
+        </record>
```
carmenbianca: dedent ;)

carmenbianca (changes requested):
Mostly LGTM, but some criticisms.

Can we try to make the compute method work?

member_card/demo/cooperators.xml
victor-champonnois: Because the customer field does not exist in 13 anymore

victor-champonnois:
Note : we spent some time investigating with Hugues on how to make the compute method work. It's likely to be a problem related to multi-company, and the fact that the original barcode field (in odoo/base) is company_dependent. 
If we use a "new"  barcode field (with another name) that is not overiding the original barcode field it works well.
We tried adding `  @api.depends_context('company')`   on the compute method, and `company_dependent=False` on the parameters. But none works. 
To be continued...

victor-champonnois:
@carmenbianca @huguesdk How do we move on with this ? Should we give up trying to make the compute method work and do it another way ?

huguesdk:
i think we should first make some tests with a minimal module to understand how we’re supposed to do it (i did something similar to test out how to do company-dependent related fields for cooperator).

carmenbianca:
I'm making some progress on this, but I am (partially) blocked by https://github.com/odoo/odoo/issues/137141

carmenbianca:
Reverted the workaround, added a fix. Setting `company_dependent=False` was enough, in spite of your prior research. Not sure how that went wrong.

Awaiting review.

member_card/demo/cooperators.xml
```diff
@@ -6,7 +6,6 @@
 <odoo>
     <record id="res_partner_cooperator_1_demo" model="res.partner">
```
robinkeunen: We should use demo data from the standard modules rather than adding new fake partners.

member_card/models/partner.py
```diff
@@ -7,14 +7,31 @@
...
+        # on this field.
+        company_dependent=False,
+        store=True,
+    )
     member_card_ids = fields.One2many("member.card", "partner_id")
 
     member_card_to_be_printed = fields.Boolean("Print Member card?")
     last_printed = fields.Datetime("Last printed on")
 
+    # TODO: Remove after https://github.com/odoo/odoo/issues/137141 is resolved.
```
robinkeunen: It is resolved.

robinkeunen (changes requested):
We're getting there

member_card/report/member_card_template.xml
robinkeunen: it is the way

robinkeunen (approved):

polchampion:
Functional test on 04/01/2023 and 12/01/2024 on [16-test-member-card](https://odoo16.main.test.srv.coopiteasy.be/web?db=16-test-member-card): 
@victor-champonnois @robinkeunen I improved the list of features a tad:

- [x] Adds a member_card_template view and print option on partner
- [x] Adds a "Member card" tab on the partner with a button to create a new member card
- [x] The partner's card and barcode history is visible in the member card tab
- [x] Creating a card generates a barcode, witch is then displayed on the member card
- [x] The "Force Barcode" option allows to set a specific barcode instead
- [x] A partner's barcode is computed from the last active member card 
- [x] Adds a field `member_card_logo` on the company allowing to upload an image
- [x] The card template displays the `member_card_logo`  image
- [x] A boolean "Print Member card?" allows to flag partners for whom you need to print new cards.
- [x] The wizards "Request member card printing" and "Set member card as printed" allow to mass check and uncheck the "Print Member Card?" flag.
- [x] If the point of sale is installed, the generated barcode matches customer pattern rule.

robinkeunen:
@carmenbianca ready for another review

polchampion:
@robinkeunen functional tests ok, I updated my previous post!

huguesdk (approved):
lgtm. i squashed the migration-related commits together.

PR #526, opened 2023-10-09
[16.0][MIG] website_portal_restrict_modification

carmenbianca:
## Description

Not sure if this really belongsin the obeesdoo repository anymore. Not sure where else to put it. Because it's not configurable, I don't think we should move this to the OCA.

## Odoo task (if applicable)

https://gestion.coopiteasy.be/web#id=10787&model=project.task&view_type=form&menu_id=

## Checklist before approval


website_portal_restrict_modification/controllers/portal.py
```diff
@@ -0,0 +1,42 @@
...
+
+        # Therefore : bring back to instance scope if not already
+        if "MANDATORY_BILLING_FIELDS" not in vars(self):
+            self.MANDATORY_BILLING_FIELDS = (
+                CustomerPortal.MANDATORY_BILLING_FIELDS.copy()
+            )
+        if "OPTIONAL_BILLING_FIELDS" not in vars(self):
+            self.OPTIONAL_BILLING_FIELDS = CustomerPortal.OPTIONAL_BILLING_FIELDS.copy()
+
+        # move name and email to optional
```
carmenbianca, line 27: Why? Isn't it safe to keep these mandatory, even if they're readonly? They seem mandatory on `res.users` _anyway_.

robinkeunen: I think it's because when they're marked as "readlonly", the fields are not sent back to the controller. Then the controller complains fields are missing.

robinkeunen (approved):
The code is technically ok for me but I'm challenging the need for it w/ Pol. cf the task.

robinkeunen:
Closing this
- the need is not clear
- the implementation is complex and bug prone 
  - class variable with server-wide scope
  - marking mandatory fields as optional fields so that the controller does not complain when displayed mandatory field are readlonly in the page
- Bart (bees and also their _bureau des membres_), Virginie (CIE) agree to remove it.

It does not need migration script if we need to reintroduce it in the future.

more info : https://gestion.coopiteasy.be/web#id=10787&action=475&active_id=492&model=project.task&view_type=form&menu_id=

PR #528, opened 2023-10-10
16.0 mig cooperator worker

robinkeunen:
## Odoo task (if applicable)

https://gestion.coopiteasy.be/web#id=10786&action=475&active_id=469&model=project.task&view_type=form&menu_id=536

## Checklist before approval


cooperator_worker/models/res_partner.py
```diff
@@ -43,6 +43,7 @@ def _compute_is_worker(self):
         """
         for rec in self:
             share_type = rec._cooperator_share_type()
+            # rec.is_worker = share_type and share_type.allow_working
```
carmenbianca: This appears added by error.

cooperator_worker/tests/test_base.py
carmenbianca, line 1: Rename to common.py

test-requirements.txt
```diff
@@ -0,0 +1 @@
+git+https://github.com/coopiteasy/cooperative@16.0-mig-cooperator-static-test-helpers#subdirectory=setup/cooperator
```
carmenbianca: Make sure to delete this commit before merging.

carmenbianca (changes requested):
Mostly LGTM, some comments.

In the future, it's easier (and standard OCA process) to create a separate migration commit for every version through which you are migrating. You don't need to test every version; the process is a development aid and helps future Git archaeologists figure out why some changes were made.

robinkeunen:
So if I understand correctly, here, I should do 
- a 13 commit changing the version
- a 14 commit changing the version
- a 15 commit adapting the tests
- a 16 commit changing the version

Correct ?

test-requirements.txt
robinkeunen: @carmenbianca are you sure I should remove this commit before merging ? The corresponding PR is still open https://github.com/OCA/cooperative/pull/86 

(test-requirements here was updated)

carmenbianca:
Basically yes @ your question on commits.

test-requirements.txt
carmenbianca: If you want this merged sooner rather than later, it's OK to merge as-is. I didn't realise this wasn't an OCA repo.

It's still not ideal (we have to remember to remove that line again), but it'll do.

carmenbianca (approved):

polchampion:
functionaly tested on 16-test-cooperator

- [x] Configure on share product whether its owner can work and shop
- [ ] Partner fields is_worker and can_shop are computed from the share they subscribed to. **NOK for can_shop, which is initially true on the partner even when it is false on the share product.** However, this existing behaviour is not new to 16, and has been discussed [here](https://github.com/beescoop/Obeesdoo/issues/449) and the situation would require a change:

As is : As long as a partner who is a cooperator doesn't have a working mode, `can_shop`  remains true.
To be in cooperator_worker : if is_worker is true, can_shop should only depend on the shift status. Else, the partner's can_shop should match the share product. Also, is the share product allows to work, it should also always allow to shop.

@carmenbianca I created an internal task to improve this.

polchampion:
Also @carmenbianca I realise that this module is really a `shift` module, since it is a glue module that adds `shift` features to `cooperator` . Should it be renamed `shift_cooperator_worker` ?

PR #535, opened 2024-10-21
[12.0][FIX] shift: Recompute partner's can_shop when dependencies change

carmenbianca:
## Description

Because `can_shop` is a stored computed field, this was previously broken. It wouldn't change, because `cooperative_status_ids` always remains identical---a single record.

## Odoo task (if applicable)

T13157

## Checklist before approval

### DIFF
--- a/shift/models/res_partner.py
+++ b/shift/models/res_partner.py
@@ -88,7 +88,7 @@ class ResPartner(models.Model):
         store=True,
     )
 
-    @api.depends("cooperative_status_ids")
+    @api.depends("cooperative_status_ids.can_shop")
     def _compute_can_shop(self):
         """
         Shopping authorisation may vary on the can_shop status of the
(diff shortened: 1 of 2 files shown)


shift/models/res_partner.py
```diff
@@ -88,7 +88,10 @@ class ResPartner(models.Model):
         store=True,
     )
 
-    @api.depends("cooperative_status_ids")
+    @api.depends(
+        "cooperative_status_ids",
+        "cooperative_status_ids.can_shop",
+    )
```
huguesdk:
```suggestion
    @api.depends("cooperative_status_ids.can_shop")
```

should be enough, no?

carmenbianca: agreed, fixed, will fix in other pr too

huguesdk (approved):

PR #546, opened 2025-05-05
[16.0][MIG] portal_eater: Migration to 16.0

mihien:


portal_eater/views/portal_eater_templates.xml
```diff
@@ -7,15 +7,15 @@
...
-        id="website_eater_details_side"
-        name="Website Eater Details Side"
+        id="portal_eater_details_side"
+        name="Portal Eater Details Side"
         priority="50"
         inherit_id="portal.portal_layout"
     >
         <xpath expr="//div[hasclass('o_portal_my_details')]" position="after">
-            <div class="o_my_details_eater mt-2">
+            <div class="o_my_details_eater mt-3">
```
huguesdk: the sections from `portal` use a class beginning with `o_portal_my_*` see [here](https://github.com/odoo/odoo/blob/16.0/addons/portal/views/portal_templates.xml#L150). i would be better to use the same prefix.

portal_eater/views/portal_eater_templates.xml
```diff
@@ -7,15 +7,15 @@
...
                 <h4 class="page-header">Your Eaters</h4>
```
huguesdk: the `page-header` class doesn’t exist anymore.

huguesdk (changes requested):
@mihien thanks for your first pr!

lgtm overall, just some small changes needed.

portal_eater/views/portal_eater_templates.xml
huguesdk: the idea was not to remove the `<div>`, but simply rename `o_my_details_eater` to `o_portal_my_eater`.

huguesdk (approved):
lgtm! 🎉 thanks @mihien!

PR #549, opened 2025-05-31
[12.0][ADD] Add a button to redirect to the webshop in the menu

mihien:
[T11126](https://gestion.coopiteasy.be/web#id=11126&action=475&active_id=467&model=project.task&view_type=form&menu_id=536)


huguesdk (changes requested):
thanks for this dev, @mihien!

i’ve made some changes directly to win some time.

this is very specific to bees coop (as their webshop url is hardcoded), while this repository contains mostly generic code that can be reused. moreover, it isn’t related to the built-in odoo webshop (which is confusing). so i think that this module should go in [`cie-custom`](https://github.com/coopiteasy/cie-custom), and be named `bees_custom_webshop_button`.

mihien:
Moved to [here](https://github.com/coopiteasy/cie-custom/pull/133)

PR #568, opened 2025-11-27
[16.0][FIX] beesdoo_product_label: change logo field from binary to image

mihien:
By default in 16 a binary field is stored as an attachment. So, after a migration, binary data that used to be in the model's table in previous versions will not be accessible anymore as the ORM will look them up in ir.attachment instead.

### DIFF
--- a/beesdoo_product_label/migrations/16.0.1.0.0/post-convert_logo_field.py
+++ b/beesdoo_product_label/migrations/16.0.1.0.0/post-convert_logo_field.py
@@ -1,0 +1,34 @@
+# SPDX-FileCopyrightText: 2026 Coop IT Easy SC
+#
+# SPDX-License-Identifier: AGPL-3.0-or-later
+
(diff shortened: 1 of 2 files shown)


huguesdk (changes requested):
good catch! i would approve like this, but the question is now: why do we want to store it in the database table instead of as an `ir.attachment`? maybe this was not intentional. this was done in #236.

after checking in odoo's code the use cases for `attachment=False`, it is only for computed files (like a qr code). so this probably needs a migration script instead of this change.

edit: moreover, it needs to become a `fields.Image` field instead of `fields.Binary`.

huguesdk:
i converted the field to a `fields.Image` and added a migration script for `16.0.1.0.0`, so that it will only run during migrations to 16.0, but not for new databases in 16.0 that already installed the module.

huguesdk (approved):

PR #569, opened 2025-12-08
[16.0][ADD] volunteer: add generators and subscriptions automation

aydrpm:
## Description

This PR adds automations for generators and subscriptions:
- Add shifts generation when generator is confirmed
- Add participation generation for new subscriptions
- Add subscriptions modification management, with cancellation and generation of participation if needed
- Manage cancellation of punctual participation if covered by a subscription (new or modified)
- Prevent generator creation in canceled state
- Add generator cancellation automation
- Add related tests and dedicated common setup for generators and subscriptions

## Checklist before approval

### DIFF
--- a/volunteer/models/volunteer_shift_recurrent_generator.py
+++ b/volunteer/models/volunteer_shift_recurrent_generator.py
@@ -81,12 +81,14 @@ class VolunteerShiftRecurrentGenerator(models.Model):
     ]
 
     # Override methods
+
     @api.model_create_multi
     def create(self, vals_list):
         generators = super().create(vals_list)
         for generator in generators:
             if generator.state == "confirmed":
                 generator._generate_shifts()
+                generator._generate_all_participation()
         return generators
 
     def write(self, vals):
@@ -94,9 +96,10 @@ class VolunteerShiftRecurrentGenerator(models.Model):
         res = super().write(vals)
         for generator in self:
             previous_state = old_states[generator.id]
-            # If the generator is confirmed, generate shifts
+            # If the generator is confirmed, generate shifts and participation
             if previous_state == "draft" and generator.state == "confirmed":
                 generator._generate_shifts()
+                generator._generate_all_participation()
         return res
 
     # Methods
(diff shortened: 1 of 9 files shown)


volunteer/models/volunteer_shift_recurrent_generator.py
```diff
@@ -93,3 +119,110 @@ def _get_interval_delta(self):
...
+                    "end_time": shift_generated_end_time,
+                    "max_volunteer_nb": self.max_volunteer_nb,
+                    "stage_id": stage_confirmed.id,
+                    "type_id": self.type_id.id,
+                    "category_id": self.category_id.id,
+                    "tag_ids": self.tag_ids.ids,
+                    "generator_id": self.id,
+                    "company_id": self.company_id.id,
+                    "coordinator_id": self.coordinator_id.id,
+                }
```
remytms, line 196: It’s a good idea to move the creation of the `vals` for the `create` function to be set in a separated function. So that it can be easily extended.

volunteer/models/volunteer_shift_recurrent_subscription.py
```diff
@@ -43,3 +43,195 @@ class VolunteerShiftRecurrentSubscription(models.Model):
...
+        self._cancel_conflicting_punctual_participation(shifts)
+        participation_to_create = []
+        for shift in shifts:
+            participation_to_create.append(
+                {
+                    "shift_id": shift.id,
+                    "volunteer_id": self.volunteer_id.id,
+                    "registration_type": "recurrent",
+                    "registration_state": "confirmed",
+                }
```
remytms, line 130: Should be nice to have a separate function for setting values for the shifts, so that it can be extended easily. But this is not a blocking point.

volunteer/models/volunteer_shift_recurrent_subscription.py
```diff
@@ -43,3 +43,195 @@ class VolunteerShiftRecurrentSubscription(models.Model):
...
+        """
+        self.ensure_one()
+        future_shifts = self.generator_id._get_future_shifts()
+        if not future_shifts:
+            return
+        new_end_date = self.end_date
+        new_start_date = self.start_date
+        date_last_shift = future_shifts.sorted("end_time", reverse=True)[
+            0
+        ].end_time.date()
```
remytms, line 195: Sometimes for performance it’s better to use search with parameters to get element sorted from the DB. Because sorting in python is slower than in C in the DB engine.

remytms (approved):

aydrpm:
Thanks for the review and all these useful comments. I’ll take them into consideration for the future.

PR #570, opened 2025-12-19
[16.0][FIX] volunteer: handle multi-day shifts in subscription management

aydrpm:
## Description

This PR fixes multi-day shift handling in subscription management by  using the shift's start_date instead of end_date as reference. This  ensures multi-day shifts are included when they start within the  subscription period, even if they extend beyond.

Additionally, the definition of "future shifts" is corrected to exclude today and start from tomorrow.

## Checklist before approval


remytms (approved):
As usual, very good code. Thanks a lot.

PR #571, opened 2025-12-23
[12.0][FIX] cooperator_eater: don't send mail if validation isn't succesfull

mihien:
[15101](https://gestion.coopiteasy.be/web#id=15101&action=479&model=project.task&view_type=form&menu_id=536)

### DIFF
--- a/cooperator_eater/models/subscription_request.py
+++ b/cooperator_eater/models/subscription_request.py
@@ -20,6 +20,14 @@ class SubscriptionRequest(models.Model):
     def validate_subscription_request(self):
         self.ensure_one()
 
+        if self.partner_id:
+            vals = self.get_eater_vals(self.partner_id, self.share_product_id)
+            # settattr will change the values of the partner object without
+            # applying the changes in the db, so we can see if the new value of
+            # the field eater is valid, without the risk of changing the result of
+            # super().validate_subscription_request() which depends on that field
+            setattr(self.partner_id, "eater", vals["eater"])
+
         invoice = super().validate_subscription_request()[0]
         partner = invoice.partner_id
 


cooperator_eater/models/subscription_request.py
```diff
@@ -20,8 +20,13 @@ def get_eater_vals(self, partner, share_product_id):
     def validate_subscription_request(self):
         self.ensure_one()
 
+        # ensure partner can be changed to be a worker before both validating
+        # the subscription and writing the partner as worker in db
+        partner = self.partner_id
+        if partner:
+            setattr(partner, "eater", "worker_eater")
```
remytms: Why not using `partner.eater = "worker_eater"` ?

cooperator_eater/models/subscription_request.py
```diff
@@ -20,8 +20,13 @@ def get_eater_vals(self, partner, share_product_id):
...
+
         invoice = super().validate_subscription_request()[0]
-        partner = invoice.partner_id
```
remytms, line 24: Don’t remove this line. If `self.partner_id` is None then the invoice partner is another one (a found one or a newly created one).

Look [here](https://github.com/OCA/cooperative/blob/12.0/cooperator/models/subscription_request.py#L708-L739).

cooperator_eater/models/subscription_request.py
```diff
@@ -20,8 +20,13 @@ def get_eater_vals(self, partner, share_product_id):
...
 
         vals = self.get_eater_vals(partner, self.share_product_id)
         partner.write(vals)
```
remytms: I think that this is not correct. We should look more deeply into the parent method.

`self.partner_id` is used only if not None.

As far as I know, the issue depends on the share product and the partner. The issue happen only if `self.partner_id` is not null and if the return of the method `get_eater_vals()` is "worker_eater" and `self.partner_id.parent_eater_id` is not null.

I would suggest something like that:


```suggestion
        # if existing partner, ensure eater settings can be changed
        # before validating the subscription request.
        if self.partner_id:
            vals = self.get_eater_vals(partner, self.share_product_id)
            self.partner_id.write(vals)

        invoice = super().validate_subscription_request()[0]
        
        partner = invoice.partner_id
        if partner != self.partner_id:
            vals = self.get_eater_vals(partner, self.share_product_id)
            partner.write(vals)
```

remytms (changes requested):
Change needed.

cooperator_eater/models/subscription_request.py
mihien: If I understood/remember everything correctly, the problem was that `partner.eater = "worker_eater"` makes the changes directly in the db (and maybe trigger onchange behaviours, but I'm not sure anymore if that's not the case with setattr as well), which we don't want to happen before validating the subscription request. Using setattr willonly change the value in the current object and will raise an error if the new value is impossible, without doing any actual change in db

cooperator_eater/models/subscription_request.py
mihien: Oops yes that's my bad, I don't think that one was intended

remytms (approved):
LGTM. Thanks for the changes.

PR #573, opened 2026-01-27
[16.0] [ADD] volunteer_attendance

Genepi314:
## Checklist before approval


volunteer_attendance/models/volunteer_shift.py
```diff
@@ -0,0 +1,33 @@
...
+    @api.depends("volunteer_participation_ids.attendance_status_id")
+    def _compute_attendance_state(self):
+        for shift in self:
+            all_attendance_status = []
+            for participation in shift.volunteer_participation_ids:
+                all_attendance_status.append(participation.attendance_status_id.name)
+            if all(all_attendance_status):
+                shift.attendance_state = "validated"
+            else:
+                shift.attendance_state = "waiting"
```
remytms, line 39: Can you try this:


```suggestion
            if all(shift.volunteer_participation_ids.attendance_status_id):
                shift.attendance_state = "validated"
            else:
                shift.attendance_state = "waiting"
```

And if it does not work, try this:


```suggestion
            if all(shift.volunteer_participation_ids.mapped("attendance_status_id")):
                shift.attendance_state = "validated"
            else:
                shift.attendance_state = "waiting"
```

volunteer_attendance/views/volunteer_shift_participation_views.xml
```diff
@@ -0,0 +1,26 @@
...
+            name="inherit_id"
+            ref="volunteer.volunteer_shift_participation_view_tree"
+        />
+        <field name="arch" type="xml">
+            <field name="cancellation_date" position="after">
+                <field name="attendance_status_id" options="{'no_create':True}" />
+            </field>
+            <field name="attendance_status_id" position="after">
+                <field name="attendance_date" />
+            </field>
```
remytms, line 21: You can group field you add if its a the same place.
```suggestion
            <field name="cancellation_date" position="after">
                <field name="attendance_status_id" options="{'no_create':True}" />
                <field name="attendance_date" />
            </field>
```

volunteer_attendance/views/volunteer_shift_views.xml
```diff
@@ -0,0 +1,62 @@
...
+                        type="object"
+                        class="oe_stat_button"
+                        icon="fa-users"
+                    >
+                        <field
+                            name="attendance_state"
+                            widget="badge"
+                            decoration-success="attendance_state == 'validated'"
+                            decoration-warning="attendance_state == 'waiting'"
+                            options="{'no_create':True}"
```
remytms: The "options" here seams not necessary to me. Have you tried without ?

volunteer_attendance/models/volunteer_shift.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def action_display_attendance_state(self):
+        """Do nothing"""
+        return
```
remytms: Here we should do a link to the participations that have no attendance status yet.

volunteer_attendance/views/volunteer_volunteer_views.xml
```diff
@@ -0,0 +1,21 @@
...
+        <field name="name">Volunteer: add attendance status for participation</field>
+        <field name="model">volunteer.volunteer</field>
+        <field name="inherit_id" ref="volunteer.volunteer_volunteer_view_form" />
+        <field name="arch" type="xml">
+            <field name="cancellation_date" position="after">
+                <field name="attendance_status_id" options="{'no_create':True}" />
+            </field>
+            <field name="attendance_status_id" position="after">
+                <field name="attendance_date" />
+            </field>
```
remytms, line 16:
```suggestion
            <field name="cancellation_date" position="after">
                <field name="attendance_status_id" options="{'no_create':True}" />
                <field name="attendance_date" />
            </field>
```

volunteer_attendance/__manifest__.py
```diff
@@ -0,0 +1,29 @@
...
+        "views/volunteer_shift_views.xml",
+        "views/volunteer_volunteer_views.xml",
+        "views/volunteer_shift_participation_views.xml",
+    ],
+    "demo": [],
+    "assets": {
+        "web.assets_backend": [
+            "volunteer/static/src/scss/volunteer_shift.scss",
+        ],
+    },
```
remytms, line 28: This should be removed.

remytms (changes requested):
Great job ! Some changes required. :)

volunteer_attendance/__manifest__.py
Genepi314: Thank you for the review ! I will make the changes soon and come back to you when it is done.

volunteer_attendance/models/volunteer_shift.py
Genepi314: The mapped function doesn't return None values, therefore I always get attendance_state = "validated" with this code. I couldn't get _compute_attendance_state() to work with a mapped list, so I kept the little loop.

volunteer_attendance/views/volunteer_shift_participation_views.xml
Genepi314: Ok ! Done.

volunteer_attendance/views/volunteer_shift_views.xml
Genepi314: Absolutely. I removed it.

volunteer_attendance/views/volunteer_volunteer_views.xml
Genepi314: Ok !

volunteer_attendance/models/volunteer_shift.py
Genepi314: Do you mean a link to the participations of the shift itself, or to all participations (of all shifts) that have no attendance status yet ? 

For the shift itself, the participations are visible and modifiable on the same view.

If I make a link to the participation view, the whole view is readonly, so no change can be made.

Genepi314: I made the link to filtered participation view, with only participations with no attendance status of the shift itself. I hope I understood your demand well, otherwise tell me if changes are still needed :-)

volunteer_attendance/models/volunteer_shift.py
remytms: Ho, indeed, `mapped` filters empty recordset and duplicate ones. Can you put a small comment to explain that ?

remytms (changes requested):
Need an explanation, but the rest is ok

volunteer_attendance/models/volunteer_shift.py
Genepi314: I can do that for sure. I've also thought of another solution to use .mapped() instead of loops. How about comparing the values this way:
```
if (len(shift.volunteer_participation_ids.mapped("attendance_status_id")) == len(shift.mapped("volunteer_participation_ids"))):
    shift.attendance_state = "validated"
```
?
Didn't try yet but thought it could work. Would that version be better than the loops above ?

Genepi314: Need to add also verification for canceled participations

Genepi314: My suggestion actually still doesn't work because the mapped function doesn't return several times the same value either. I left a comment above _cancel_holiday_shift stating why I used loops instead of mapped and filtered.

Here is my non-operational attempt. It doesn't work because attendance status aren't unique to each participation:
``` python
    @api.depends("volunteer_participation_ids.attendance_status_id")
    def _compute_attendance_state(self):
        for shift in self:
            confirmed_participations = shift.volunteer_participation_ids.filtered(
                lambda participation: participation.registration_state == "confirmed"
            )
            if (
                len(set(confirmed_participations.mapped("attendance_status_id"))) 
                == len(confirmed_participations)
            ):
                shift.attendance_state = "validated"
            else:
                shift.attendance_state = "waiting"
```
Hence I modified the loops above so that they only check confirmed participations.

remytms: with the comment, it’s clearer.

remytms (approved):
LGTM.

PR #576, opened 2026-02-19
[16.0][FIX] pos_eater: linter errors

remytms:
It tries to solve theses linter errors:

pos_eater/static/src/js/ActionpadWidget.esm.js
  7:1   warning  Imports should be sorted alphabetically                                      sort-imports
  9:31  warning  'ActionpadWidget' is already declared in the upper scope on line 7 column 8  no-shadow

## Checklist before approval

### DIFF
--- a/pos_eater/static/src/js/ActionpadWidget.esm.js
+++ b/pos_eater/static/src/js/ActionpadWidget.esm.js
@@ -3,10 +3,10 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-import Registries from "point_of_sale.Registries";
 import ActionpadWidget from "point_of_sale.ActionpadWidget";
+import Registries from "point_of_sale.Registries";
 
-const EaterActionpadWidget = (ActionpadWidget) =>
+const EaterActionpadWidget = () =>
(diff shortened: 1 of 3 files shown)


pos_eater/static/src/js/ActionpadWidget.esm.js
```diff
@@ -3,10 +3,10 @@
 //
 // SPDX-License-Identifier: AGPL-3.0-or-later
 
-import Registries from "point_of_sale.Registries";
 import ActionpadWidget from "point_of_sale.ActionpadWidget";
+import Registries from "point_of_sale.Registries";
 
-const EaterActionpadWidget = (ActionpadWidget) =>
+const EaterActionpadWidget = () =>
```
huguesdk, line 9: i think that this will not work anymore. because the `class extends ActionpadWidget` line is supposed to extend the class that is received as argument. you can suffix it (in both places thus) with an underscore to avoid a name clash. example [here](https://github.com/OCA/pos/blob/16.0/pos_tare/static/src/js/Screens/ScaleScreen/ScaleScreen.esm.js#L9).

PR #578, opened 2026-03-02
[16.0] [ADD] volunteer_holiday

Genepi314:
## Description

Add holidays for companies and leaves for volunteers.

## Checklist before approval


volunteer_holiday/models/volunteer_company_holiday.py
```diff
@@ -0,0 +1,114 @@
...
+                        {
+                            "stage_id": self.env.ref(
+                                "volunteer.volunteer_shift_stage_canceled"
+                            ).id
+                        }
+                    )
+        # Back to original user
+        self.env = previous_env
+
+    def _shift_covers_holiday(
```
remytms, line 106: Add `@api.model` above function definition, as this method does not use data comming from the instance.


```suggestion
    @api.model
    def _shift_covers_holiday(
```

volunteer_holiday/models/volunteer_company_holiday.py
```diff
@@ -0,0 +1,114 @@
...
+            or (
+                shift_end_date >= holiday_start_date
+                and shift_start_date <= holiday_end_date
+            )
+            or (
+                shift_start_date <= holiday_start_date
+                and shift_end_date >= holiday_end_date
+            )
+        ):
+            return True
```
remytms: Simply write:
```suggestion
        return (
            (
                shift_start_date >= holiday_start_date
                and shift_start_date <= holiday_end_date
            )
            or (
                shift_end_date >= holiday_start_date
                and shift_start_date <= holiday_end_date
            )
            or (
                shift_start_date <= holiday_start_date
                and shift_end_date >= holiday_end_date
            )
        )
```

volunteer_holiday/models/volunteer_company_holiday.py
Genepi314: Ok !

volunteer_holiday/models/volunteer_company_holiday.py
Genepi314: Ok, I didn't know about this simplified syntax. It is done.

volunteer_holiday/controllers/__init__.py
```diff
@@ -0,0 +1 @@
+from . import controllers
```
remytms: You should remove the `controllers` directory if your have nothing more than the `__init__.py` file in it.

volunteer_holiday/data/cron.xml
```diff
@@ -0,0 +1,31 @@
...
+    <record id="ir_cron_cancel_holiday_shift" model="ir.cron">
+        <field name="name">Cancel Generated Shifts During Company Holidays</field>
+        <field name="model_id" ref="model_volunteer_company_holiday" />
+        <field name="state">code</field>
+        <field name="code">model._cancel_holiday_shift()</field>
+        <field name="interval_number">24</field>
+        <field name="interval_type">hours</field>
+        <field name="numbercall">-1</field>
+        <field name="doall" eval="False" />
+    </record>
```
remytms: Apply your comment here. :)

volunteer_holiday/demo/volunteer_company_holiday_demo.xml
```diff
@@ -0,0 +1,37 @@
...
+-->
+<odoo>
+    <record
+        id="volunteer_company_holiday_demo_christmas"
+        model="volunteer.company.holiday"
+    >
+        <field name="name">Christmas Holidays</field>
+        <field
+            name="start_date"
+            eval="(datetime.today().replace(month=12, day=24)).strftime('%Y-%m-%d')"
```
remytms: Starting with python 3.9 (odoo 16.0 uses 3.10), prefer using `isformat()` instead of strftime for standard iso formatting.

See method [here](https://docs.python.org/3.10/library/datetime.html?highlight=datetime#datetime.date.isoformat).

volunteer_holiday/demo/volunteer_company_holiday_demo.xml
```diff
@@ -0,0 +1,37 @@
...
+        />
+        <field
+            name="end_date"
+            eval="(datetime.today().replace(month=12, day=25)).strftime('%Y-%m-%d')"
```
remytms: isoformat()

volunteer_holiday/demo/volunteer_company_holiday_demo.xml
```diff
@@ -0,0 +1,37 @@
...
+    </record>
+
+    <record
+        id="volunteer_company_holiday_demo_store_inventory"
+        model="volunteer.company.holiday"
+    >
+        <field name="name">Store Inventory</field>
+        <field
+            name="start_date"
+            eval="(datetime.today().replace(month=7, day=1)).strftime('%Y-%m-%d')"
```
remytms: isoformat

volunteer_holiday/demo/volunteer_company_holiday_demo.xml
```diff
@@ -0,0 +1,37 @@
...
+        />
+        <field
+            name="end_date"
+            eval="(datetime.today().replace(month=7, day=7)).strftime('%Y-%m-%d')"
```
remytms: isoformat

volunteer_holiday/demo/volunteer_volunteer_leave_demo.xml
```diff
@@ -0,0 +1,42 @@
...
+<odoo>
+    <record
+        id="volunteer_volunteer_leave_demo_vol_a_leave"
+        model="volunteer.volunteer.leave"
+    >
+        <field name="volunteer_id" ref="volunteer.volunteer_demo_a" />
+        <field name="type_id" ref="volunteer_volunteer_leave_type_demo_vacation" />
+        <field
+            name="start_date"
+            eval="(datetime.now().date() + timedelta(weeks=4, days=4)).strftime('%Y-%m-%d')"
```
remytms: isoformat

volunteer_holiday/demo/volunteer_volunteer_leave_demo.xml
```diff
@@ -0,0 +1,42 @@
...
+        />
+        <field
+            name="end_date"
+            eval="(datetime.now().date() + timedelta(weeks=4, days=14)).strftime('%Y-%m-%d')"
```
remytms: isoformat

volunteer_holiday/demo/volunteer_volunteer_leave_demo.xml
```diff
@@ -0,0 +1,42 @@
...
+        model="volunteer.volunteer.leave"
+    >
+        <field name="volunteer_id" ref="volunteer.volunteer_demo_c" />
+        <field
+            name="type_id"
+            ref="volunteer_volunteer_leave_type_demo_parental_leave"
+        />
+        <field
+            name="start_date"
+            eval="(datetime.now().date() + timedelta(days=4)).strftime('%Y-%m-%d')"
```
remytms: isoformat

volunteer_holiday/demo/volunteer_volunteer_leave_demo.xml
```diff
@@ -0,0 +1,42 @@
...
+        />
+        <field
+            name="end_date"
+            eval="(datetime.now().date() + timedelta(weeks=9, days=4)).strftime('%Y-%m-%d')"
```
remytms: isoformat

volunteer_holiday/models/volunteer_company_holiday.py
```diff
@@ -0,0 +1,111 @@
...
+
+from dateutil.relativedelta import relativedelta
+
+from odoo import api, fields, models
+
+
+class VolunteerCompanyHoliday(models.Model):
+    _name = "volunteer.company.holiday"
+    _description = "Company Holidays"
+    _order = "start_date"
```
remytms:
```suggestion
    _order = "start_date desc, company_id, id"
```
Setting a default `_order` has a cost and it default to `id`. But I understand that most recent holliday should be shown on top (so use `desc`).

Also, `start_date` is not a unique field (several records can have the same value), so you need to specify other fields to order them correctly. Here in my suggestion, `company_id` is not needed, but `id` at the end is mandatory to be sure that we can correctly order all element.

volunteer_holiday/models/volunteer_company_holiday.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+
+    start_date = fields.Date(required=True)
+
+    end_date = fields.Date(required=True)
+
+    # Methods
+
+    # Retirer la limite de temps, regarder les shifts futurs
+    def _cancel_holiday_shift(self, time_in_months=3):
```
remytms: I think that you should apply your comment, and remove the time limit. Just check all shifts, and see if they are overlapping a holiday.

remytms: I think that for now it’s ok if the method starts at `now()` and all future shifts.

Because, we don’t want to change the past.

volunteer_holiday/models/volunteer_company_holiday.py
```diff
@@ -0,0 +1,111 @@
...
+                    ("start_date", ">=", today_midnight),
+                    ("start_date", "<=", date_time_range.date()),
+                ]
+            )
+        )
+
+        # Create dictionary key-company for list of values-shifts
+        shifts_by_company = {}
+        for shift in confirmed_future_generated_shifts_in_range:
+            shifts_by_company.setdefault(shift.company_id, []).append(shift)
```
remytms: Just order them by `company_id` when using the `search()` method.

Also, as there is less holiday than shifts, it’s better to do so with holiday, or even fetch holiday for the company when needed.

volunteer_holiday/models/volunteer_company_holiday.py
```diff
@@ -0,0 +1,111 @@
...
+                    holiday.start_date,
+                    holiday.end_date,
+                ):
+                    shift.sudo().write(
+                        {
+                            "stage_id": self.env.ref(
+                                "volunteer.volunteer_shift_stage_canceled"
+                            ).id
+                        }
+                    )
```
remytms: If your shifts are ordered by `company_id`, you can use `itertools.groupby()` to iterate through all the shifts, company by company. Then you can fetch the holiday of the comany, each time you face a new company.

Using `itertools.groupby()` is more efficient for such a task.

volunteer_holiday/models/volunteer_volunteer.py
```diff
@@ -0,0 +1,15 @@
...
+from odoo import fields, models
+
+
+class VolunteerVolunteer(models.Model):
+    _inherit = "volunteer.volunteer"
+
+    volunteer_leave_ids = fields.One2many(
+        comodel_name="volunteer.volunteer.leave",
+        inverse_name="volunteer_id",
+        string="Leave",
```
remytms: Here for the user interface you can use plural form.
```suggestion
        string="Leaves",
```

volunteer_holiday/models/volunteer_volunteer_leave.py
```diff
@@ -0,0 +1,93 @@
...
+
+from datetime import date, datetime
+
+from odoo import api, fields, models
+
+
+class VolunteerVolunteerLeave(models.Model):
+    _name = "volunteer.volunteer.leave"
+    _description = "Volunteer Leave"
+    _order = "start_date"
```
remytms:
```suggestion
    _order = "start_date desc, volunteer_id, id"
```
Same as for company.holiday.

remytms (changes requested):
First partial review.

volunteer_holiday/demo/volunteer_company_holiday_demo.xml
aydrpm: After some tests on these cases, neither `strftime()` nor `isoformat()` seems to be required. On my side, loading works fine without any formatting. That said, I'm not sure of all the details behind it.

volunteer_holiday/models/volunteer_company_holiday.py
Genepi314: Just corrected it in the the commit "Refactor _cancel_holiday_shift..."

volunteer_holiday/models/volunteer_company_holiday.py
Genepi314: I used itertools as you suggested recently and ordered the holidays by company. I'm not sure I did it perfectly right though, so don't hesitate to tell me if something is still off.

volunteer_holiday/models/volunteer_company_holiday.py
Genepi314: See comment above.

volunteer_holiday/models/volunteer_volunteer_leave.py
Genepi314: I just did this for volunteer leaves. But I am not sure I understand what you mean for company holidays. Do you mean you also want it to be displayed with _order = "start_date desc"  ? Or do you mean I should also add  "company_id, id" ?  

I didn't put the "desc" after _order = "start_date" because of the difference in the views. Volunteer Leaves has a view that I couldn't get filtered (the one nested in volunteer_volunteer_view, as a page in the notebook), so I thought it would be less tedious to get the most distant in the future as first to appear on the view, rather than the most distant in the past. 

This does not apply to company holidays, as they're filtered automatically to show only the future ones. Therefore I thought it would be best to see the closest next one, rather than the most distant in the future one. 

But anyway, this is the way I see it. If you don't like it I'll change it to "start_date desc" for company holidays too :-)

volunteer_holiday/demo/volunteer_company_holiday_demo.xml
Genepi314: > After some tests on these cases, neither `strftime()` nor `isoformat()` seems to be required. On my side, loading works fine without any formatting. That said, I'm not sure of all the details behind it.

Indeed, it is not needed. Thank you, I removed the `strftime()` in all demo files :-)

volunteer_holiday/demo/volunteer_company_holiday_demo.xml
Genepi314: I removed all `strftime()` after aydrpm's suggestion :-)

volunteer_holiday/demo/volunteer_company_holiday_demo.xml
Genepi314: I removed all `strftime()` after aydrpm's suggestion :-)

volunteer_holiday/models/volunteer_company_holiday.py
remytms: up for this comment.

volunteer_holiday/models/volunteer_company_holiday.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+
+        future_company_holidays = self.env["volunteer.company.holiday"].search(
+            [("start_date", ">=", today_midnight)]
+        )
+
+        for company, grouped_holidays_by_company in groupby(
+            future_company_holidays, key=lambda holiday: holiday.company_id.id
+        ):
+            grouped_holidays = list(grouped_holidays_by_company)
```
remytms: `grouped_holiday_by_company` is a iterator. You loop only once over it. So there is no need to convert it to a list.

volunteer_holiday/models/volunteer_company_holiday.py
```diff
@@ -0,0 +1,126 @@
...
+            hour=0, minute=0, second=0, microsecond=0
+        )
+
+        future_company_holidays = self.env["volunteer.company.holiday"].search(
+            [("start_date", ">=", today_midnight)]
+        )
+
+        for company, grouped_holidays_by_company in groupby(
+            future_company_holidays, key=lambda holiday: holiday.company_id.id
+        ):
```
remytms:
```suggestion
        for company_id, grouped_holidays_by_company in groupby(
            future_company_holidays, key=lambda holiday: holiday.company_id
        ):
```
I found this more readable in terms of Odoo nomenclature.

volunteer_holiday/models/volunteer_shift.py
```diff
@@ -0,0 +1,47 @@
...
+                    shift.start_time,
+                    shift.end_time,
+                    holiday.start_date,
+                    holiday.end_date,
+                ):
+                    found_overlap = True
+                    shift.overlaps_holiday = True
+
+            if not found_overlap:
+                shift.overlaps_holiday = False
```
remytms: Here you should not retrieve all shift and then check them. The compute function must only run on record in `self`.

So just group self by company_id. Then loop over the shift and fetch company holiday when needed.

Then you must set a value for `overlaps_holiday` for each record in `self`.

volunteer_holiday/models/volunteer_volunteer.py
```diff
@@ -0,0 +1,38 @@
...
+        inverse_name="volunteer_id",
+        string="Leaves",
+    )
+
+    def _send_notification_end_leave(self):
+        """Send a notification to volunteers before their leave ends."""
+        all_companies = self.env["res.company"].search([])
+        for company in all_companies:
+            nb_days = company.nb_days_before_leave_end
+            message_body = ("Your leave ends in {} days.").format(nb_days)
```
remytms: Here import the `_` function and use it for you string like this:

`_("Your ...")`

This way, your string can be translated.

volunteer_holiday/models/volunteer_volunteer_leave.py
```diff
@@ -0,0 +1,78 @@
...
+        """Cancel participations of volunteers if they overlap with their time off"""
+        Holiday = self.env["volunteer.company.holiday"]
+        today_midnight = datetime.today().replace(
+            hour=0, minute=0, second=0, microsecond=0
+        )
+        all_future_volunteer_leaves = (
+            self.env["volunteer.volunteer.leave"]
+            .sudo()
+            .search([("end_date", ">=", date.today())])
+        )
```
remytms: I wonder if `sudo()` is needed, as we can run the cron with a user that has the correct rights to do that.

volunteer_holiday/models/volunteer_volunteer_leave.py
```diff
@@ -0,0 +1,78 @@
...
+                )
+            )
+            for participation in future_confirmed_participations:
+                if Holiday._shift_covers_holiday(
+                    participation.shift_id.start_time,
+                    participation.shift_id.end_time,
+                    leave.start_date,
+                    leave.end_date,
+                ):
+                    participation.sudo().write({"registration_state": "canceled"})
```
remytms: Idem here for `sudo()`.

volunteer_holiday/models/volunteer_volunteer_leave_type.py
```diff
@@ -0,0 +1,21 @@
...
+    _order = "name"
+
+    name = fields.Char(string="Leave Type", required="True")
+    description = fields.Char()
+
+    company_id = fields.Many2one(
+        comodel_name="res.company",
+        string="Company",
+        default=lambda self: self.env.user.company_id,
+        required=True,
```
remytms: Here we can remove the `required=True` for company_id. As if it is false it means that it can be used by all company.

remytms (changes requested):
Partial review.

volunteer_holiday/models/volunteer_company_holiday.py
Genepi314: I see, it is changed.

volunteer_holiday/models/volunteer_volunteer_leave.py
Genepi314: True, I changed it in company holidays but not in volunteer leaves. Thanks for telling me.

volunteer_holiday/models/volunteer_volunteer_leave_type.py
Genepi314: I removed it there. 

But would that be true for all company_id fields then ? Like in volunteer.company.holiday for example ? Or is it different for this one ? I don't know how it would actually work with the cancel_holiday_shift function, as for now the domain used to retrieve shifts from the same company has a `"="` operator for companies, and not a `"in"`...

volunteer_holiday/models/volunteer_shift.py
Genepi314: I wrote this one a bit quickly, so sorry for the lack of optimization already. 
I have some questions here:
1) Should I actually use an iterator like in company_holidays to go through holidays and shifts more efficiently ? 
2) About setting a value for overlaps_holiday, I thought it wasn't needed, as it has a default value set to false already. Should I still set it to false again in the compute function ?

volunteer_holiday/data/cron.xml
```diff
@@ -0,0 +1,48 @@
...
+            name="name"
+        >Notification: Send a Reminder to Volunteers Before their Leave Ends</field>
+        <field name="model_id" ref="model_volunteer_volunteer" />
+        <field name="state">code</field>
+        <field name="code">model._send_notification_end_leave()</field>
+        <field name="interval_number">24</field>
+        <field name="interval_type">hours</field>
+        <field name="numbercall">-1</field>
+        <field name="doall" eval="False" />
+    </record>
```
remytms, line 46: Here we should set the admin user as `user_id`.

volunteer_holiday/models/volunteer_company_holiday.py
Genepi314: Ok, thanks. I left start_date with no desc, because otherwise you would see company holidays in one year before the ones in one month for example. But I changed company_id and id.

volunteer_holiday/__manifest__.py
```diff
@@ -0,0 +1,34 @@
...
+    "name": "Volunteer Holiday",
+    "summary": "Add holidays for companies and volunteers",
+    "version": "16.0.0.5.0",
+    "category": "Volunteer management",
+    "website": "https://github.com/beescoop/Obeesdoo",
+    "author": "Coop IT Easy SC",
+    "maintainers": ["remytms"],
+    "license": "AGPL-3",
+    "application": False,
+    "depends": ["base", "volunteer", "mail"],
```
remytms, line 15:
```suggestion
    "depends": ["volunteer", "mail"],
```

volunteer_holiday/views/volunteer_volunteer_leave_view.xml
```diff
@@ -0,0 +1,98 @@
...
+SPDX-FileCopyrightText: 2025 Coop IT Easy SC
+
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<odoo>
+    <record id="volunteer_volunteer_leave_view_tree" model="ir.ui.view">
+        <field name="name">Volunteer Leaves List</field>
+        <field name="model">volunteer.volunteer.leave</field>
+        <field name="arch" type="xml">
+            <tree default_order="start_date asc">
```
remytms, line 12: No need to put default_order if it is not different than the one in `_order` of the model.

remytms (changes requested):
Some little changes.

volunteer_holiday/views/volunteer_volunteer_leave_view.xml
Genepi314: Here it is different in the model and in the view. Here is the code in the model:
`_order = "start_date desc, volunteer_id, id"`
This is because of the notebook page in volunteer_volunteer_view in which no default filter is available, where as in the volunteer_volunteer_leave_view, there is a default filter applied to see only future volunteer leaves.

PR #579, opened 2026-03-02
[16.0][ADD] volunteer: Add generator write permissions

aydrpm:
## Description

This PR adds role-based write permissions for generators, covering field modifications, state transitions, and subscription management via the notebook.

State transition rules:
- Cannot return to draft from any state
- Cannot change state of a canceled generator
- Cannot modify other fields or subscriptions while canceling

Role-based field permissions:
- User: readonly via access rules
- Manager: 
	- cannot modify fields
	- cannot change state
	- can only manage subscriptions (create/update/cancel) on draft and confirmed generators
- Admin: 
	- can modify fields on draft generators
	- can change state when allowed
	- can manage subscriptions on draft and confirmed generators

Blocked for all on canceled generators:
- No field modification
- No subscription management

UI:
- Subscription cancel button hidden for user group
- Notebook readonly for canceled generator
- Procedure guidance message on confirmed generator field modification

Tests files added:
- test_volunteer_shift_recurrent_generator_state_change_permissions
- test_volunteer_shift_recurrent_generator_write_permissions

## Checklist before approval

### DIFF
--- a/volunteer/models/volunteer_shift_recurrent_generator.py
+++ b/volunteer/models/volunteer_shift_recurrent_generator.py
@@ -7,7 +7,7 @@ from datetime import timedelta
 from dateutil.relativedelta import relativedelta
 
 from odoo import api, fields, models
-from odoo.exceptions import UserError, ValidationError
+from odoo.exceptions import AccessError, UserError, ValidationError
 from odoo.tools.translate import _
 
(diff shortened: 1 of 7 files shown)


remytms (approved):
LGTM!

PR #580, opened 2026-03-04
[16.0][FIX] volunteer_attendance: various fixes

Genepi314:
…ity file with record rule for company access on volunteer shift attendance status.

## Checklist before approval


volunteer_attendance/models/volunteer_shift_attendance_status.py
```diff
@@ -19,7 +19,7 @@ class VolunteerShiftAttendanceStatus(models.Model):
         comodel_name="res.company",
         string="Company",
         default=lambda self: self.env.user.company_id,
-        required=True,
+        # required=True,
```
remytms: Either leave required or remove but do not comment. I think here it should be removed as a False value for Company is ok.

remytms (changes requested):
One small change. Then it is ok.

volunteer_attendance/models/volunteer_shift_attendance_status.py
Genepi314: Ah sorry, didn't remove comments yet. I'll trim it now :-)

Genepi314: Done.

volunteer_attendance/models/volunteer_shift_participation.py
```diff
@@ -35,8 +35,8 @@ def _check_no_attendance_status_for_canceled_participation(self):
...
     # Methods
 
-    @api.depends("attendance_status_id")
     def write(self, vals):
         """Set attendance status modification date to now at modification"""
-        vals["attendance_date"] = fields.Datetime.now()
-        return super().write(vals)
+        if "attendance_status_id" in vals:
+            vals["attendance_date"] = fields.Datetime.now()
+            return super().write(vals)
```
remytms: Ho, this one, I didn’t catch it the first time.

write should always return and call the parent function. So write this:

```suggestion
        if "attendance_status_id" in vals:
            vals["attendance_date"] = fields.Datetime.now()
        return super().write(vals)
```

remytms (changes requested):
Still one change. Sorry I should have seen it the first time.

volunteer_attendance/models/volunteer_shift_participation.py
Genepi314: Oh, missed it myself. Thanks, done !

Genepi314:
Just fixed README so that Remy is now visible as maintainer.

volunteer_attendance/models/volunteer_shift_participation.py
```diff
@@ -23,20 +23,19 @@ def _check_no_attendance_status_for_canceled_participation(self):
...
-                and participation.attendance_status_id is not None
+                and participation.attendance_status_id.id
             ):
                 raise ValidationError(
                     _(
                         "An attendance status for a canceled participation "
-                        "will not be taken into account for the attendance state "
-                        "of the shift."
+                        "will not be taken into account. "
                     )
```
remytms: I want to suggest this:

"You can’t set attendance status on a cancelled participation."

volunteer_attendance/readme/ROADMAP.rst
```diff
@@ -0,0 +1,10 @@
+- The purpose of the smart button used for the attendance state in the shift form view is not clear.
+
+    - Redesign the smart button
+
+    - A help attribute for the attendance state field could help understand better this functionality
```
remytms: - count the number of participation without an attendance state.

volunteer_attendance/__manifest__.py
```diff
@@ -9,21 +9,20 @@
     "category": "Volunteer management",
     "website": "https://github.com/beescoop/Obeesdoo",
     "author": "Coop IT Easy SC",
+    "maintainers": ["remytms"],
     "license": "AGPL-3",
     "application": False,
     "depends": ["base", "volunteer", "mail"],
```
remytms:
```suggestion
    "depends": ["volunteer", "mail"],
```
All modules depends on "base".

remytms (changes requested):
Some small changes. Otherwise, very good. :)

volunteer_attendance/models/volunteer_shift_participation.py
Genepi314: It is more direct. Changing it now.

remytms (approved):
Yeah!

PR #581, opened 2026-03-05
[12.0] [FIX] shift: ordering

remytms:
## Description

Models that have an `_order` parameter are not configured correctly.

There were configured only with fields that are not unique for the
model.

For example shift.template had `_order = "start_time"`. start_time
is not a unique field for shift.template. You can have several
shift.template with the same value for start_time.

If you fetch directly some shift.template, there is no issue. You get
shift.template ordered by start_time. And items with the same value for
start_time are returned in a random order. You can use the UI to sort or
filter them with other fields.

The issue occurs when you try to fetch some shift.shift and you want
them ordered by shift.template.
```python
shifts = self.env["shift.shift"].search([], _order="shift_template_id")
```
You assume that shifts will be grouped by shift.template. But as soon as
you have two shift.template with the same start_time, this is not the
case.

Odoo uses the _order specified on the model shift.template to
order shift.template in the request of shift.shift.

You get something suprising like this:
```
shift_id | shift_template_id | start_time
---------|-------------------|-----------
1        | 1                 | 10
2        | 2                 | 10
3        | 1                 | 10
4        | 1                 | 10
5        | 2                 | 10
6        | 3                 | 14
7        | 3                 | 14
```

In the example all the shift.template that have 10 as start_time are
considered equal. Therefore shift are not ordered by shift.template.

What we expect is the following result:
```
shift_id | shift_template_id | start_time
---------|-------------------|-----------
1        | 1                 | 10
3        | 1                 | 10
4        | 1                 | 10
2        | 2                 | 10
5        | 2                 | 10
6        | 3                 | 14
7        | 3                 | 14
```

To get this result it is important to set at least and in final position
a unique field in the _order attribute of the model. For example the id.

With `_order = "start_time, id"` in the model shift.template, we get the
right result.

This fix is applied to other models than the one exposed in this
example.

## Odoo task (if applicable)

[task](https://gestion.coopiteasy.be/web#id=15900&action=479&model=project.task&view_type=form&menu_id=536)

## Checklist before approval

### DIFF
--- a/shift/models/cooperative_status.py
+++ b/shift/models/cooperative_status.py
@@ -25,7 +25,7 @@ class HistoryStatus(models.Model):
     _name = "cooperative.status.history"
     _description = "cooperative.status.history"
 
-    _order = "create_date desc"
+    _order = "create_date desc, cooperator_id, id"
 
     status_id = fields.Many2one("cooperative.status")
     cooperator_id = fields.Many2one("res.partner")
@@ -40,7 +40,7 @@ class CooperativeStatus(models.Model):
     _name = "cooperative.status"
     _description = "cooperative.status"
     _rec_name = "cooperator_id"
-    _order = "cooperator_id"
+    _order = "cooperator_id, id"
     _period = 28
 
     def _get_status(self):
@@ -505,7 +505,7 @@ class CooperativeStatus(models.Model):
 class ShiftCronJournal(models.Model):
     _name = "shift.journal"
     _description = "shift.journal"
-    _order = "date desc"
+    _order = "date desc, id"
     _rec_name = "date"
 
     date = fields.Date()
--- a/shift/models/planning.py
+++ b/shift/models/planning.py
@@ -31,6 +31,8 @@ class ShiftType(models.Model):
     _name = "shift.type"
     _description = "shift.type"
 
+    _order = "name, id"
+
     name = fields.Char()
     description = fields.Text()
     active = fields.Boolean(default=True)
@@ -40,7 +42,7 @@ class ShiftDayNumber(models.Model):
     _name = "shift.daynumber"
     _description = "shift.daynumber"
 
-    _order = "number asc"
+    _order = "number asc, id"
 
     name = fields.Char()
     number = fields.Integer(
@@ -55,7 +57,7 @@ class ShiftDayNumber(models.Model):
 class ShiftPlanning(models.Model):
     _name = "shift.planning"
     _description = "shift.planning"
-    _order = "sequence asc"
+    _order = "sequence asc, id"
 
     sequence = fields.Integer()
     name = fields.Char()
@@ -198,7 +200,7 @@ class ShiftPlanning(models.Model):
 class ShiftTemplate(models.Model):
     _name = "shift.template"
     _description = "shift.template"
-    _order = "start_time"
+    _order = "planning_id, task_type_id, start_time, id"
 
     name = fields.Char(required=True)
     planning_id = fields.Many2one("shift.planning", required=True)
--- a/shift/models/shift_shift.py
+++ b/shift/models/shift_shift.py
@@ -10,7 +10,7 @@ from odoo.tools.translate import _
 class ShiftShift(models.Model):
     _name = "shift.shift"
     _inherit = ["mail.thread"]
-    _order = "start_time asc"
+    _order = "start_time asc, task_template_id, task_type_id, worker_id, id"
 
     ##################################
     # Method to override             #
--- a/shift/__manifest__.py
+++ b/shift/__manifest__.py
@@ -14,7 +14,7 @@
     ),
     "website": "https://github.com/beescoop/Obeesdoo",
     "category": "Cooperative management",
-    "version": "12.0.5.0.2",
+    "version": "12.0.5.0.3",
     "depends": ["mail"],
     "data": [
         "data/system_parameter.xml",
--- a/shift_attendance/tests/test_shift.py
+++ b/shift_attendance/tests/test_shift.py
@@ -336,6 +336,8 @@ class TestShift(TransactionCase):
         self.assertEqual(sheet_1.added_shift_ids[1].task_id.state, "done")
 
         # Empty shift should have been updated
-        self.assertEqual(sheet_1.added_shift_ids[0].task_id, self.shift_empty_1)
+        self.assertIn(
+            self.shift_empty_1, (rec.task_id for rec in sheet_1.added_shift_ids)
+        )
 
         # sheet_1.expected_shift_ids[0].worker_id


shift/models/planning.py
```diff
@@ -198,7 +198,7 @@ def get_future_shifts(
 class ShiftTemplate(models.Model):
     _name = "shift.template"
     _description = "shift.template"
-    _order = "start_time"
+    _order = "planning_id, task_type_id, start_time, id"
```
huguesdk, line 201: is it intentional to heavily change the order here?

huguesdk (commented):
i’m not sure i understand: is the problem due to the fact that the order of records that have the same value for the field used in `_order` is different from time to time, and you want to ensure that it always stays the same? are you using id to have an extra value meaning “creation order”? i think that the commit message should be more explicit about the “why”, because i think that this sentence is not correct:

> Setting only field that are not unique in the "_order" attribute is not correct.

or maybe i am missing something?

shift/models/planning.py
remytms: Yes, it is intentional. Ordering by "start_time" (knowing that start time is only a float value representing hours) has no sens.

remytms:
@huguesdk I edited the message to have a better explanation of the issue. Is it clearer now ?

shift/models/cooperative_status.py
```diff
@@ -25,7 +25,7 @@ class HistoryStatus(models.Model):
     _name = "cooperative.status.history"
     _description = "cooperative.status.history"
 
-    _order = "create_date desc"
+    _order = "create_date desc, id"
```
huguesdk: i think that ordering by `id` is too random here. `cooperator_id`, `user_id` or `status_id` (the one that makes the most sense) will work too (as they all use an order that is based on a unique field).

shift/models/shift_shift.py
```diff
@@ -10,7 +10,7 @@
 class ShiftShift(models.Model):
     _name = "shift.shift"
     _inherit = ["mail.thread"]
-    _order = "start_time asc"
+    _order = "start_time asc, task_template_id, task_type_id, worker_id, id"
```
huguesdk, line 13: not sure that `id` is needed here, as all the other fields make for a unique condition (but it doesn’t hurt anyway). `worker_id` uses a unique field to order itself. one thing to fix, though, is that `shift.type` (used by `task_type_id`) has no `_order`; it should be `"name, id"` (or `name` should be unique).

huguesdk (changes requested):
thanks a lot for the very clear explanation in the commit message. i think a short explanation just above the `_order` definitions in the code explaining the why would also be useful (and more visible than in a commit message). something like:
```python
    # id is needed to ensure uniqueness, which ensures that other models
    # ordered by this model have their records grouped by these records.
```

also, `shift.type` needs a unique `_order` condition.

shift/models/cooperative_status.py
remytms: I can not replace `id` by `cooperator_id` because `cooperator_id` (well defined and distinguishable by id) is not a unique field on this model.

Nothing prevent you from having two cooperative.status.history to belong to the same cooperator_id idem for user_id and idem for status_id.

So I can add other fields and then ends with `id`. But I can not remove `id` as it is the only unique field of this model.

shift/models/shift_shift.py
remytms: It’s not a matter of whether the fields can be distinguish by it’s own _order values. We should ensure that, in all _order attribute, there is at least one field with a unique constraint. ID always have, sometimes we can use another field if there is such a unique constraint.

shift/models/cooperative_status.py
huguesdk: ok, but then only using `id` is too random: the order will make no sense. there should be other fields used before `id`.

shift/models/shift_shift.py
huguesdk: are you sure of this? for example, if you order by worker_id, the results will be correctly grouped, even if there are 2 partners with the same name, because `res.partner` uses `name` and `id`.

remytms: Yes, I’m pretty sure.

Here is what I understand.

If you order `shift.shift` by `worker_id`, as res.partner has `_order = "name, id"`, `shift.shift` will be properly ordered because it will differentiate two partners with the same name.

The issue occurs if you try to order another model which have a many2one field linking to a `shift.shift`. Let’s call this model "A".

You can have several items A, that are linked to different shifts.

E.g.
```
model_a:
A_id | shift_id
-----|---------
1    | 4
2    | 3
3    | 4
```
With this :
```
shift_shift:
id | start_time | task_template_id | task_type_id | worker_id
---|------------|------------------|--------------|----------
3  | 10         | 1                | 1            | 1
4  | 10         | 1                | 1            | 1
```
(I know that in the case of shift, it’s a non sens that the same worker subscribe to two different shifts of the same task_template, but there is nothing in the code that prevent that, as far as I know, at least not with a constrain in the database).

If you order model A by shift_id, shift 3 and 4 will be equivalent because all the fields in `_order` are the same. Or these are two different shifts, there are not equal event if their value are equal.

To get this result:
```
model_a:
A_id | shift_id
-----|---------
2    | 3
1    | 4
3    | 4
```
We need to set a unique field in the `_order` attribute of the `shift.shift` model. Else you get something like the previous table for model A. Where model A linked to shift 3 or 4 are considered equal and not sorted.

shift/models/cooperative_status.py
remytms: I think that `id` is mandatory. But other fields are not. I agree that it’s better to have a default order that make sens. But I’m not sure that it’s a good idea to put lots of field for each model in terms of performances. I have no certitude about that. So I will apply the changes requested. :)

shift/models/shift_shift.py
huguesdk: thanks again for the explanation. ok, if all fields have the same value, then the only distinction between them is the `id`, indeed. i was assuming that the combination of the fields used in the `_order` property would be unique, as it should to avoid the problem you’re fixing here. but indeed, even if it makes no sense, since there is no constraint preventing this, it could happen. maybe adding a constraint could make sense here, then? i’m not against adding `id` in `_order`, as sometimes it is the only solution (like for `res.partner`, for example), but for others, i think that it would be better to think about the uniqueness constraints first and leveraging this if possible.

remytms: I think that putting `id` is a good trade of, it’s safe and not ambiguous.

We can think about adding other constraints, but I don’t find evident one.

I said in my previous message that there is no sens to have the same partner for the two shift of the same shift template. But in fact if these two shift are cancelled than it make sens to keep track of the subscription and cancellation date.So the case exposed above can exist.

Also if Odoo lets create Primary key with several fields, than it can be OK. But that’s not the case. Odoo always adds an ID field.

I fact, I think that even if there is constraints on some field, as other module can override this, it’s safer to keep ID in the _order field.

PR #574, opened 2026-02-10
[FIX] use product deposit rather than deposit tax

mihien:
## Description

use product deposit rather than deposit tax

## Odoo task (if applicable)

[12145](https://gestion.coopiteasy.be/web?debug=1#id=12136&action=479&model=project.task&view_type=form&menu_id=536)

PR #583, opened 2026-03-13
[16.0][IMP] allow to use deposit products

huguesdk:
## description

this supersedes #574 (without dropping compatibility with deposit taxes).

* add dependency to pos_container_deposit to allow to use deposit products to compute deposit price.
* simplify taxes computation.

## odoo task


mihien (approved):
Lgtm

PR #584, opened 2026-03-13
[16.0][FIX] fix display of product quantity

huguesdk:
## description

* fix display of product quantity: use correct decimal separator and uom precision.
* remove extraneous (breaking) space in unit price display.

## odoo task


mihien (approved):
Lgtm

PR #585, opened 2026-03-13
[16.0][REM] remove module beesdoo_label_deposit

huguesdk:
## description

this was a temporary module, and it is actually not used (and not working because of an error in a file path). this behavior is now handled by `sale_product_deposit` directly (see #583).

## odoo task


mihien (approved):
Lgtm

PR #586, opened 2026-03-16
[16.0][FIX] pos_eater: fix missing class argument

huguesdk:
## description

fix missing class argument in `EaterActionpadWidget` (erroneously removed in #576).


remytms (approved):
Thank you very much ! :)

PR #589, opened 2026-05-11
[16.0][ADD] shift_portal_display_attendees

mihien:
## Description

- Add shift_portal_display_attendees 
- Some UI changes in shift_portal
- Add _get_shift_coworkers in shift_portal

## Odoo task (if applicable)

[15899](https://gestion.coopiteasy.be/web?debug=1#id=15899&action=479&model=project.task&view_type=form&menu_id=536)

## Checklist before approval


shift_portal/models/website.py
```diff
@@ -48,3 +48,21 @@ class Website(models.Model):
...
+                    ("task_template_id", "=", shift.task_template_id.id),
+                    ("start_time", "=", shift.start_time),
+                    ("task_type_id", "=", shift.task_type_id.id),
+                    ("worker_id", "!=", False),
+                ]
+            )
+        )
+        return all_shifts.mapped("worker_id").filtered(
+            lambda p: p.id != self.env.user.partner_id.id
+        )
```
remytms: I think that this is not the right place for this function.

Have you think about putting that function into the `shift.shift` model?

It can also go in the controller if not possible in the `shift.shift` model.

remytms: I wonder if the domain can be written like this:


```suggestion
            .search(
                [
                    ("task_template_id", "=", shift.task_template_id.id),
                    ("start_time", "=", shift.start_time),
                    ("task_type_id", "=", shift.task_type_id.id),
                    ("worker_id", "!=", False),
                    ("worker_id", "!=", self.env.user.partner_id.id),
                ]
            )
        )
        return all_shifts.mapped("worker_id")
```

shift_portal/views/my_shift_website_templates.xml
```diff
@@ -281,26 +279,77 @@
...
+                                                    t-att-href="'mailto:%s' % shift.super_coop_id.partner_id.email"
+                                                    class="text-info"
+                                                    style="font-size:12px;"
+                                                >
+                                                    <i class="fa fa-envelope me-1" />
+                                                    <t
+                                                        t-out="shift.super_coop_id.partner_id.email"
+                                                    />
+                                                </a>
+                                            </t>
```
remytms: Put the if directly in the `a` tag to avoid unneeded indentation.

shift_portal/views/my_shift_website_templates.xml
```diff
@@ -281,26 +279,77 @@
...
+                                                    t-att-href="'tel:%s' % shift.super_coop_id.partner_id.phone"
+                                                    class="text-info"
+                                                    style="font-size:12px;"
+                                                >
+                                                    <i class="fa fa-phone me-1" />
+                                                    <t
+                                                        t-out="shift.super_coop_id.partner_id.phone"
+                                                    />
+                                                </a>
+                                            </t>
```
remytms: Same about the `t-if` in the `a` tag.

shift_portal/views/my_shift_website_templates.xml
```diff
@@ -454,22 +503,51 @@
...
+                        <div class="modal-body">
+    <div class="d-grid gap-2" style="grid-template-columns: 1fr 1fr;">
+
+        <div class="d-flex align-items-center gap-3 p-2 rounded bg-light">
+            <i class="fa fa-calendar text-muted" />
+            <div>
+                <small
+                                        class="text-uppercase text-muted"
+                                        style="font-size:10px; letter-spacing:0.04em;"
+                                    >Date &amp; time</small>
```
remytms: This is strangely indented.

shift_portal/views/my_shift_website_templates.xml
```diff
@@ -454,22 +503,51 @@
...
-                    </div>
+                        <div class="modal-body">
+    <div class="d-grid gap-2" style="grid-template-columns: 1fr 1fr;">
+
+        <div class="d-flex align-items-center gap-3 p-2 rounded bg-light">
+            <i class="fa fa-calendar text-muted" />
+            <div>
+                <small
+                                        class="text-uppercase text-muted"
+                                        style="font-size:10px; letter-spacing:0.04em;"
```
remytms: Do not put `style` attribute. Use defaults from bootstrap, or create a CSS file.

remytms: So, here, try to juste remove the `style` attribute and see if it is still readable.

shift_portal/views/my_shift_website_templates.xml
```diff
@@ -454,22 +503,51 @@
...
+                </div>
+            </div>
+        </div>
+
+        <div class="d-flex align-items-center gap-3 p-2 rounded bg-light">
+            <i class="oi oi-studio text-muted" />
+            <div>
+                <small
+                                        class="text-uppercase text-muted"
+                                        style="font-size:10px; letter-spacing:0.04em;"
```
remytms: Same for `style` attribute.

shift_portal/views/my_shift_website_templates.xml
```diff
@@ -454,22 +503,51 @@
...
+                <div><t t-out="shift.task_type_id.name" /></div>
+            </div>
+        </div>
+
+        <div class="d-flex align-items-center gap-3 p-2 rounded bg-light">
+            <i class="fa fa-user-plus text-muted" />
+            <div>
+                <small
+                                        class="text-uppercase text-muted"
+                                        style="font-size:10px; letter-spacing:0.04em;"
```
remytms: Same for style attribute.

shift_portal/views/my_shift_website_templates.xml
```diff
@@ -532,18 +611,60 @@
...
+                            style="grid-template-columns: 1fr 1fr;"
+                        >
+                            <div
+                                class="d-flex align-items-center gap-3 p-2 rounded bg-light"
+                            >
+                                <i class="fa fa-calendar text-muted" />
+                                <div>
+                                    <small
+                                        class="text-uppercase text-muted"
+                                        style="font-size:10px; letter-spacing:0.04em;"
```
remytms: Same for `style` attribute.

shift_portal/views/my_shift_website_templates.xml
```diff
@@ -532,18 +611,60 @@
...
+                            </div>
+
+                            <div
+                                class="d-flex align-items-center gap-3 p-2 rounded bg-light"
+                            >
+                                <i class="oi oi-studio text-muted" />
+                                <div>
+                                    <small
+                                        class="text-uppercase text-muted"
+                                        style="font-size:10px; letter-spacing:0.04em;"
```
remytms: Same for `style` attribute.

shift_portal/views/my_shift_website_templates.xml
```diff
@@ -532,18 +611,60 @@
...
+                            </div>
+
+                            <div
+                                class="d-flex align-items-center gap-3 p-2 rounded bg-light"
+                            >
+                                <i class="fa fa-user-plus text-muted" />
+                                <div>
+                                    <small
+                                        class="text-uppercase text-muted"
+                                        style="font-size:10px; letter-spacing:0.04em;"
```
remytms: Same for `style` attribute.

shift_portal_display_attendees/controllers/main.py
```diff
@@ -0,0 +1,28 @@
...
+
+class WebsiteShiftController(WebsiteShiftController):
+    def available_shift_irregular_worker(
+        self, irregular_enable_sign_up=False, nexturl=""
+    ):
+        res = super().available_shift_irregular_worker(
+            irregular_enable_sign_up, nexturl
+        )
+        res["display_attendees"] = request.website.display_attendees_on_portal
+        return res
```
remytms: I think that `request.website.display_attendees_on_portal` is directly accessible from the template. So is it really needed to override `available_shift_irregular_worker` ?

remytms: It is better in fact ton not compute values in the view. But here there is no computation at all, just assigning a variable to another. So maybe it too much code for nothing.

shift_portal_display_attendees/controllers/main.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def get_compensation_shift_grid(self, shifts):
+        res = super().get_compensation_shift_grid(shifts)
+        res["display_attendees"] = request.website.display_attendees_on_portal
+        return res
+
+    def my_shift_next_shifts(self, partner=None):
+        res = super().my_shift_next_shifts(partner)
+        res["display_attendees"] = request.website.display_attendees_on_portal
+        return res
```
remytms: Same comment as above.

shift_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,172 @@
...
+            expr="//div[hasclass('modal-body')]//div[hasclass('d-grid')]"
+            position="inside"
+        >
+        <t t-if="shift.super_coop_id">
+            <div class="d-flex align-items-start gap-3 p-2 rounded bg-light">
+                <i class="fa fa-user-circle text-muted mt-1" />
+                <div>
+                    <small
+                            class="text-uppercase text-muted"
+                            style="font-size:10px; letter-spacing:0.04em;"
```
remytms: Same comment for `style` attribute.

shift_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,172 @@
...
+                        <t t-if="shift.super_coop_id.partner_id.phone">
+                            <a
+                                    t-att-href="'tel:%s' % shift.super_coop_id.partner_id.phone"
+                                    class="text-info"
+                                    style="font-size:12px;"
+                                >
+                                <i class="fa fa-phone me-1" />
+                                <t t-out="shift.super_coop_id.partner_id.phone" />
+                            </a>
+                        </t>
```
remytms: Same comment for the `t-if` in the `a` tag.

shift_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,172 @@
...
+            <t t-set="workers" t-value="request.website._get_shift_coworkers(shift)" />
+            <div
+                    class="d-flex align-items-start gap-3 p-2 rounded bg-light"
+                    style="grid-column: 1 / -1;"
+                >
+                <i class="fa fa-users text-muted mt-1" />
+                <div>
+                    <small
+                            class="text-uppercase text-muted"
+                            style="font-size:10px; letter-spacing:0.04em;"
```
remytms: Same comment for `style` attribute.

shift_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,172 @@
...
+                        >Super cooperator</small>
+                        <div><t t-out="shift.super_coop_id.name" /></div>
+                        <div class="d-flex flex-column gap-1 mt-1">
+                            <t t-if="shift.super_coop_id.partner_id.email">
+                                <a
+                                    t-att-href="'mailto:%s' % shift.super_coop_id.partner_id.email"
+                                    class="text-info"
+                                    style="font-size:12px;"
```
remytms: Same comment for `style` attribute.

shift_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,172 @@
...
+                                >
+                                <i class="fa fa-envelope me-1" />
+                                <t t-out="shift.super_coop_id.partner_id.email" />
+                            </a>
+                        </t>
+                        <t t-if="shift.super_coop_id.partner_id.phone">
+                            <a
+                                    t-att-href="'tel:%s' % shift.super_coop_id.partner_id.phone"
+                                    class="text-info"
+                                    style="font-size:12px;"
```
remytms: Same comment for `style` attribute.

shift_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,172 @@
...
+                            <t t-if="shift.super_coop_id.partner_id.phone">
+                                <a
+                                    t-att-href="'tel:%s' % shift.super_coop_id.partner_id.phone"
+                                    class="text-info"
+                                    style="font-size:12px;"
+                                >
+                                    <i class="fa fa-phone me-1" />
+                                    <t t-out="shift.super_coop_id.partner_id.phone" />
+                                </a>
+                            </t>
```
remytms: Same comment for `t-if` in `a` tag.

shift_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,172 @@
...
+            </t>
+
+            <t t-if="display_attendees and request.website._get_shift_coworkers(shift)">
+                <t
+                    t-set="workers"
+                    t-value="request.website._get_shift_coworkers(shift)"
+                />
+                <div
+                    class="d-flex align-items-start gap-3 p-2 rounded bg-light"
+                    style="grid-column: 1 / -1;"
```
remytms: Here, not sure what does the `grid-column`. Also should not be in a `style` attribute. Maybe it’s not needed, try without. Or put it in a CSS file.

shift_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,172 @@
...
+                >
+                    <i class="fa fa-users text-muted mt-1" />
+                    <div>
+                        <small
+                            class="text-uppercase text-muted"
+                            style="font-size:10px; letter-spacing:0.04em;"
```
remytms: Same comment for `style` attribute.

shift_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,172 @@
...
+            position="inside"
+        >
+            <t t-if="display_attendees and request.website._get_shift_coworkers(shift)">
+                <t
+                    t-set="workers"
+                    t-value="request.website._get_shift_coworkers(shift)"
+                />
+                <div
+                    class="d-flex align-items-start gap-3 p-2 rounded bg-light"
+                    style="grid-column: 1 / -1;"
```
remytms: Same comment for `style` attribute.

shift_portal_display_attendees/views/templates.xml
remytms, line 1: There is way to much duplicated code here in this file. Use templates with variable to factorize you code and no rewrite the same thing everywhere.

shift_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,172 @@
...
+                                    <i class="fa fa-phone me-1" />
+                                    <t t-out="shift.super_coop_id.partner_id.phone" />
+                                </a>
+                            </t>
+                        </div>
+                    </div>
+                </div>
+            </t>
+
+            <t t-if="display_attendees and request.website._get_shift_coworkers(shift)">
```
remytms: Here for example. Getting value for `display_attendeees` directly from `request.website` make sens. But getting coworkers is a work for the controller. Here this function is called several times. Call it once. Either with a `t-set` or in the controller or directly on the shift. Depending on where your putting the function `_get_shift_coworkers`.

remytms (changes requested):
There is too many duplicated code. Also a function can be moved elsewhere.

shift/models/shift_shift.py
```diff
@@ -398,6 +398,24 @@ def _cron_send_weekly_emails(self, notice=1, period=7):
         for rec in confirmed_tasks:
             shift_summary_mail_template.send_mail(rec.id, True)
 
+    def _get_shift_coworkers(self):
+        """Get coworkers for a shift (workers registered for the same task template,
+        start time and task type)."""
+        all_shifts = (
+            self.env["shift.shift"]
```
remytms:
```suggestion
            self
```
No need to call `env["shift.shift"]` because `self` is already a recordset of type "shift.shift".

shift/models/shift_shift.py
```diff
@@ -398,6 +398,24 @@ def _cron_send_weekly_emails(self, notice=1, period=7):
...
+            .sudo()
+            .search(
+                [
+                    ("task_template_id", "=", self.task_template_id.id),
+                    ("start_time", "=", self.start_time),
+                    ("task_type_id", "=", self.task_type_id.id),
+                    ("worker_id", "!=", False),
+                    ("worker_id", "!=", self.env.user.partner_id.id),
```
remytms:
```suggestion
                    ("worker_id", "!=", self.worker_id.id),
```
Isn't it better to exclude the worker of this shift. So that we get the co-workers of this shift ?

Or maybe this function should just return all worker of this slot, without excluding any.

I just found a bit weird that this function response depends on the user connected.

So for me two options:
1. not filtering (except false value)
2. filtering the worker of self.

If option 1. then you can filter in the view like:
```
<t t-out="worker.name" t-if="worker != request.env.user.partner_id" />
```

If option 2. then you should be sure that `self` is the shift with `worker_id` equal to the current connected user.

I think I prefer option 1.

remytms (changes requested):
Some change requested.

But very good job in template refactoring. It’s way way better like this. :)

shift/models/shift_shift.py
mihien: Oops yeah, didn't think about that when I moved it from the controller. Good catch

shift/models/shift_shift.py
mihien: I agree, I'll go with option 1

shift_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!--
+     SPDX-FileCopyrightText: 2026 Coop IT Easy SC
+     SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<odoo>
+    <template id="shift_detail_modal" inherit_id="shift_portal.shift_detail_modal">
+        <xpath expr="//div[hasclass('row')]" position="inside">
+            <t t-if="display_attendees">
```
remytms:
```suggestion
            <t t-if="request.website.display_attendees_on_portal">
```

shift_portal_display_attendees/controllers/main.py
remytms: Remove this file, and use `request.website.display_attendees_on_portal` directly in the view. :)

remytms (changes requested):
Some changes requested.

shift_change_portal/views/change_templates.xml
```diff
@@ -135,7 +135,7 @@ SPDX-License-Identifier: AGPL-3.0-or-later
             <a
                 t-attf-href="/my/shift/change/{{ old_shift.id }}/{{ shift.id }}"
                 t-if="regular_shift_change_enabled and shift.id and not is_subscribed"
-                class="btn btn-sm btn-primary m-1"
+                class="btn btn-sm btn-primary m-1 fw-bold"
```
remytms, line 138: Ho, it’s `fw` and not `font-weight` now. Good to know. :)

shift_change_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!--
+     SPDX-FileCopyrightText: 2026 Coop IT Easy SC
+     SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<odoo>
+    <!-- 1. Change the Select button from direct link to modal trigger -->
```
remytms: This comment does not seams correct.

shift_change_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,24 @@
...
+    <template
+        id="available_shift_irregular_worker_card_change_button"
+        inherit_id="shift_change_portal.available_shift_irregular_worker_card"
+    >
```
remytms, line 11: I think that this is not correct.

`shift_change_portal.available_shift_irregular_worker_card` does not exist.

I think you try to inherit from:
`shift_portal.available_shift_irregular_worker_card`

shift_change_portal_display_attendees/views/templates.xml
remytms, line 1: It’s nice to add this information button that trigger the modal.

But I’m not sure that this modal exist in all the views.

Be sure to add this modal in the `shift_portal.available_shift_irregular_worker_grid` view.

shift_portal/views/my_shift_website_templates.xml
```diff
@@ -268,7 +350,6 @@
             <!-- Super Cooperator info modal -->
```
remytms: Change comment here as it is no longer only for Super Cooperator information, but for shift details.

remytms (changes requested):
It need still some changes.

shift_change_portal_display_attendees/views/templates.xml
remytms: It works, so no change needed.

shift_change_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<!--
+     SPDX-FileCopyrightText: 2026 Coop IT Easy SC
+     SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<odoo>
+    <!-- Add Shift Details button to irregular worker card -->
+    <template
+        id="available_shift_irregular_worker_card_change_button"
```
remytms:
```suggestion
        id="available_shift_irregular_worker_card_shift_detail_button"
```

shift_change_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,29 @@
...
+        inherit_id="shift_portal.available_shift_irregular_worker_card"
+    >
+        <xpath expr="//div" position="inside">
+            <t
+                t-set="is_public_user"
+                t-value="request.env.user == request.website.user_id"
+            />
```
remytms, line 16: Very good catch this one. :)

remytms (changes requested):

remytms (approved):
LGTM.

remytms (approved):
LGTM

PR #572, opened 2026-01-09
[16.0][MIG] beesdoo_shift_swap -> shift_change

remytms:
## Description

This brings `shift_change` which brings some features of the the `beesdoo_shift_swap` module.

It keeps only the `swap` mechanism. So regular worker can only unsubscribe from a shift and subscribe to a new empty one (only in the one already generated shifts).

## Odoo task (if applicable)

[task](https://gestion.coopiteasy.be/web?debug=1#id=15056&action=479&model=project.task&view_type=form&menu_id=536)

## Checklist before approval

- [x] Tests are present (or not needed).
- [x] Credits/copyright have been changed correctly.
- [ ] Change log snippet is present.
- [x] (If a new module) Moving this to OCA has been considered.

PR #590, opened 2026-05-11
[12.0] [FIX] beesdoo_website_shift: empty shifts in portal

remytms:
## Description

Shifts is shown in case of free_space = 0 and hide_rule = 0. Which is not correct. Shift should only be shown if free_space > 0 and free_space >= special rules.

Fix in 16.0 is available in https://github.com/beescoop/Obeesdoo/pull/572.

## Odoo task (if applicable)

[task](https://gestion.coopiteasy.be/web#id=16390&action=479&model=project.task&view_type=form&menu_id=536)

## Checklist before approval

### DIFF
--- a/beesdoo_website_shift/controllers/main.py
+++ b/beesdoo_website_shift/controllers/main.py
@@ -449,7 +449,7 @@ class WebsiteShiftController(http.Controller):
             has_enough_workers = (
                 free_space <= (task_template.worker_nb * highlight_rule_pc) / 100
             )
-            if free_space >= task_template.worker_nb * hide_rule:
+            if free_space > 0 and free_space >= task_template.worker_nb * hide_rule:
                 displayed_shifts.append(
                     DisplayedShift(
                         shift_list[0],


beesdoo_website_shift/controllers/main.py
```diff
@@ -449,7 +449,7 @@ def available_shift_irregular_worker(
             has_enough_workers = (
                 free_space <= (task_template.worker_nb * highlight_rule_pc) / 100
             )
-            if free_space >= task_template.worker_nb * hide_rule:
+            if free_space and free_space >= task_template.worker_nb * hide_rule:
```
huguesdk:
```suggestion
            if free_space > 0 and free_space >= task_template.worker_nb * hide_rule:
```
explicit is better than implicit. ☺️ i think that like this it is more clear. with only `if free_space` (without knowing the code) i thought that it could also maybe be `False` or `None`.

huguesdk (approved):

beesdoo_website_shift/controllers/main.py
remytms: indeed. fixed.

PR #591, opened 2026-05-18
[16.0][ADD] shift_portal_change_display_attendees

mihien:
## Description

Based on #572 for the module shift_change and on #589 for shift_portal_display_attendees. 
This PR serves two roles:
- apply the (minimal) changes needed for shift_portal_display_attendees to work with the new changes on shift_portal
- add shift_portal_change_display_attendees, which displays the attendees of a shift on a shift change

## Odoo task (if applicable)

[15899](https://gestion.coopiteasy.be/web#id=15899&action=479&model=project.task&view_type=form&menu_id=536)

## Checklist before approval


shift_change_portal_display_attendees/controllers/main.py
```diff
@@ -0,0 +1,20 @@
...
+    def choose_new_shift(self, old_shift_id, **kw):
+        res = super().choose_new_shift(old_shift_id, **kw)
+
+        # Inject display_attendees into the template context
+        if hasattr(res, "qcontext"):
+            res.qcontext[
+                "display_attendees"
+            ] = request.website.display_attendees_on_portal
+
+        return res
```
remytms, line 20: `request.website.display_attendees_on_portal` is available directly on the view. So no need to override this method to add it.

If you need to fetch some data or need to do some computation, then it need to go in the controller. If it’s just accessing some variable then you can do it directly in the view.

shift_change_portal_display_attendees/views/templates.xml
```diff
@@ -0,0 +1,61 @@
...
+            <a
+                class="btn btn-primary"
+                t-if="irregular_enable_sign_up"
+                t-att-href="'/shift/%s/subscribe?nexturl=%s' % (shift.id, nexturl)"
+            >Subscribe</a>
+            <a
+                class="btn btn-primary"
+                t-if="regular_shift_change_enabled"
+                t-att-href="'/my/shift/change/%s/%s' % (old_shift.id, shift.id)"
+            >Subscribe</a>
```
remytms, line 57: I think that this is not correct. The regular should not be allowed to subscribe to a random shift even if "regular_shift_change_enabled" is true.

In the parent view, it’s only allowed when `irregular_enable_sign_up` is true.

shift_change_portal_display_attendees/views/templates.xml
remytms, line 1: Can’t we just add a button with "Shift detail", instead of adding modal into the workflow.

I think that this will lead to less errors or complicated code.

remytms (changes requested):
I think we need to add a "shift detail" on the portal calendar view of the shifts instead of bringing information into modal that popup in the workflow. It will be simpler.

mihien:
Now part of #589

PR #602, opened 2026-09-16
[16.0][FIX] remove duplicate shift settings menuitem

huguesdk:
## description

remove duplicate shift settings menuitem that already exists [in the shift module](https://github.com/beescoop/Obeesdoo/blob/16.0/shift/views/menu.xml#L132).

## odoo task

[t16576](https://gestion.coopiteasy.be/web#id=16576&model=project.task)


remytms (approved):
Indeed! :)

